PR #3, opened 2015-07-02
Initial report on june 2015

agjohnson:
Copy for report on June activity


agjohnson:
Looks fine to me. I originally left out minor updates, but +1 on including those

agjohnson:
New Ablog version fixed the janky encoding on RSS feed

PR #6, opened 2015-09-07
2015 State of the Docs post

agjohnson:


state-of-the-docs-2015.rst
```diff
@@ -0,0 +1,197 @@
...
+State of the Docs
+=================
+
+August has historical significance for Read the Docs, so it seemed fitting to
+wrap up August by taking a moment to step back and reflect on what we've done in
+the last year.
+
+*What makes August so significant for us?*
+
+For starters, this August saw the **5th birthday of Read the Docs**!  We're
```
ericholscher: This should probably link to the initial release post: http://ericholscher.com/blog/2010/aug/16/announcing-read-docs/

state-of-the-docs-2015.rst
```diff
@@ -0,0 +1,197 @@
...
+still humbled by the project's success, and the important role it continues to
+play for old and new users alike.
+
+This growth has been occasionally difficult to manage, especially when there
```
ericholscher: This is kind of negative, for a celebratory post.

state-of-the-docs-2015.rst
```diff
@@ -0,0 +1,197 @@
...
+
+Stability
+    Read the Docs has "grown organically", which is to say it has grown fast,
+    unwieldy, and in many directions.  With more proper focus on code quality,
+    including clean up efforts, refactoring, and long due upgrades, Read the
+    Docs has shed considerable technical debt in the last couple months.  This
+    work has already paid off in improving stability and predictability of the
+    project.
+
+Gold
```
ericholscher: We've never talked about this publicly..this seems like a weird place to introduce it?

state-of-the-docs-2015.rst
```diff
@@ -0,0 +1,197 @@
...
+    We decided to adopt a "gold" status feature model for users that want to
+    support us each month. This is an optional, recurring subscription to Read
+    the Docs.
+
+    We have plans to make gold subscriptions more functional, but for now, it is
+    just a way to show your support of Read the Docs.
+
+Ads
```
ericholscher: Similarly, odd to have this be basically a footnote in a yearly post as first introduction.

state-of-the-docs-2015.rst
```diff
@@ -0,0 +1,197 @@
...
+
+    Feedback regarding ads has been positive, hopefully our intentions with ads
+    has been mostly apparent.
+
+The next year
+-------------
+
+So, let us now set some expectations for the next year forward:
+
+Monthly updates
```
ericholscher: Whats the value of providing monthly updates? I feel like its overhead that doesn't buy us a lot -- seems like we should just be promoting new features/etc, rather than doing "updates".

state-of-the-docs-2015.rst
```diff
@@ -0,0 +1,197 @@
...
+    build backend processes closely.
+
+More reference doc language support
+    We have been working towards wider language support for generating API
+    reference documentation. Most recently, we worked with Microsoft to develop
+    support for `.NET language support in Sphinx <sphinx-dotnet_>`_. This allows
+    authors to use existing .NET tooling to generate reStructuredText API
+    reference documentation.
+
+.. _sphinx-dotnet: https://github.com/rtfd/sphinx-autoapi
```
ericholscher: Should this be a link to the Domain?

state-of-the-docs-2015.rst
agjohnson: If we want this to be more of a celebratory post, we can. I wrote this strictly as a review though, so yeah, some parts are more negative for the sake of transparency.

This could be reworded to convey that the solution to handling growth is being on the project full time, not  part time.

state-of-the-docs-2015.rst
agjohnson: As it's in production already, it seems weird not to. I'd like to follow up with a post introducing gold officially, once we're happy with where it is in terms of feature completeness. Maybe we just allude to an official introduction while still mentioning its existence here?

state-of-the-docs-2015.rst
agjohnson: Ads, or Gold? We can promise the same here, more of an introduction later, and maybe less detail here. But I think we should mention it.

state-of-the-docs-2015.rst
agjohnson: I'm pretty enthusiastic about more transparency into our decisions, which we don't offer any outlet for now. Update posts offer a good structure for setting what we're working towards and allow for some public introspection, where features posts are pure marketing.

If we don't want to commit to this at all, we should find other ways to communicate this. If we don't want to commit to this as much, we could provide quarterly updates. Monthly is probably overkill, and won't net many readers, but would be the easiest to write.

state-of-the-docs-2015.rst
ericholscher: In production -- but not promoted or having users know about.

state-of-the-docs-2015.rst
ericholscher: Both -- though the ads were a bit more prominent, so make more sense to mention.

state-of-the-docs-2015.rst
ericholscher: Yea, monthly updates are a way of communicating what we decided -- not getting feedback. If we want to be more transparent, we should be including people in the decision making process, not sharing our decisions after we've made them.

agjohnson: Getting feedback is something else to call out here. We need both in some fashion, we can't expect input without exposing more about the organization.

These posts should be just as much about what we're planning to do as they are about what we accomplished/failed at, not just enumerating our decisions.

ericholscher: Makes sense. :+1:

state-of-the-docs-2015.rst
agjohnson: This could do more to introduce the feature conceptually then, and still avoid promoting it as a usable feature. We haven't done anything to hide it thus far, I don't see mentioning it here working against promoting it later either.

PR #8, opened 2015-10-16
Added blog post about markdown support

ericholscher:


adding-markdown-support.rst
```diff
@@ -0,0 +1,74 @@
+.. post:: Oct 16, 2015
+   :tags: commonmark, markdown, feature
+
+Read the Docs & Sphinx now support Commonmark
+=============================================
+
+Read the Docs has long been built on top of Sphinx as it's primary technology.
+Sphinx for a long time has been built on top of reStructuredText as it's primary input mechanism.
```
agjohnson: These two senetences are fragmenty on their own, consider joining them. Also, "long been/time" feels unsure. "Read the Docs was built on top of Sphinx, which has always relied on rST as an input mechanism." Though maybe that's not all true

adding-markdown-support.rst
```diff
@@ -0,0 +1,74 @@
...
+
+It should be noted that Commonmark doesn't support a lot of the concepts that RST lets you represent.
+In particular,
+there is no standardized way in Commonmark to represent inline or block levels constructs.
+So things like the ``toctree`` directive and ``:ref:`` markup don't have an analog.
+
+There is a `proposed draft`_ to the Commonmark spec to allow a similar syntax.
+We are investigating adding support to this in recommonmark,
+but we don't want to implement something,
+and then have something else become the standard.
```
agjohnson: 'something/something else' could be more specific. ".. we don't want to implement a standard prematurely" or somesuch

PR #22, opened 2017-04-27
Add initial post around ad changes

ericholscher:
This is a fleshed out version of my thinking,
and would love feedback on the ideas.
In particular,
does it logically make sense?

There will be some code changes required to update this,
but I'm considering this documentation driven dev.
So once we agree on a path forward,
I can implement it.

### DIFF
--- a/ads-on-other-themes.rst
+++ b/ads-on-other-themes.rst
@@ -21,11 +21,17 @@ We've now hit that limit,
 so we're planning to expand to our second most popular theme:
 **Alabaster**
 
+We're also launching a **Community Ads** initiative,
+which will highlight projects in the community that we believe in and care about,
+and use 10% of our traffic for those ads.
+
 To explicitly repeat:
 **we will start showing ads on Alabaster themes on Read the Docs** starting in the beginning of May.
+We plan to roll this out in stages:
 
-We're also launching a **Community Ads** initiative,
-which will highlight projects in the community that we believe in and care about. 
+* We'll start with an initial ad that points to this blog post for a few days
+* Then showing our Community Ad on Alabaster projects for a few days
+* Finally, we'll expand to our full paid ad rotation
 
 Read below to learn more about these specific projects,
 and what we're doing with the money it brings in.
--- a/ads-on-other-themes.rst
+++ b/ads-on-other-themes.rst
@@ -1,8 +1,8 @@
 .. post::
    :tags: advertising, community, updates, community ads
 
-Ads on Other Themes on Read the Docs Community Sites
-====================================================
+Ads on the Alabaster Theme on Read the Docs Community Sites
+===========================================================
 
 We've been running our `ethical advertising <http://docs.readthedocs.io/en/latest/ethical-advertising.html>`_ campaign for over a year now,
 and it is starting to show success on making Read the Docs more sustainable. 
--- a/ads-on-other-themes.rst
+++ b/ads-on-other-themes.rst
@@ -53,6 +53,8 @@ There are a large number of projects that we care about in the software ecosyste
 We'll show 10% of our ad inventory each month to support an open source project that we care about.
 We are working to set up an application process to be considered,
 but we've already picked our first Community Ad recipient: **Beeware**.
+They are running their own sustainability campaign,
+and we're hoping to help out!
 
 `Beeware <http://pybee.org/>`_ is a collection of projects that can be used to help develop, debug and launch Python software. Each tool follows the Unix philosophy of doing one thing well. The primary goal is to bring Python to the browser, mobile devices, and everywhere else that we wish Python ran natively. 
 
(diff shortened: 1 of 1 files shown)


agjohnson:
The structure of this is clear and concise, it seems to work well for this post. I felt like there were opportunities to talk more about grand vision, or more about how this is all part of our plan for advertising, and not just a happy coincidence.

For instance, part of our grand vision is that the ads support community members. While we're inferring that point, I think explicitly calling that out as one of the purposes of the ad initiative is important. I don't think we need to get into our future plans at all, but perhaps mention that we do have even grander plans for this if it succeeds.

I think this makes all the points we are trying to make otherwise, and avoids getting into the weeds around contract roles, future roles, etc etc.

ads-on-other-themes.rst
```diff
@@ -0,0 +1,127 @@
...
+`Beeware <http://pybee.org/>`_ is a collection of projects that can be used to help develop, debug and launch Python software. Each tool follows the Unix philosophy of doing one thing well. The primary goal is to bring Python to the browser, mobile devices, and everywhere else that we wish Python ran natively. 
+
+`Sustain <https://sustainoss.org/>`_ is a one-day conference about open source sustainability. We ran ads for it earlier this month, and it is now sold out.
+
+Project & User opt out changes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Previously we've allowed project owners to opt out of advertising. We're now expanding this to allow users to opt out of advertising as well.
+However,
+we're using the opportunity of launching our Community Ads to be able show them to folks who out opt.
```
agjohnson: This sentence seems to be trying to say multiple things, or is ambiguous in what it is trying to say. This probably needs to be condensed to only convey that users will not be able to opt out of community ads.

ads-on-other-themes.rst
```diff
@@ -0,0 +1,127 @@
...
+We know some folks disagree with the ideas of companies paying for their attention,
+but we believe that seeing community projects promoted is something that should have value for people.
```
agjohnson: `.. that seeing community projects promoted is something that should have value for people` is passive and a bit hard to parse. An approximation of `.. that supporting the efforts of projects in the community should be important to everyone` is closer to what we're trying to convey.

ads-on-other-themes.rst
```diff
@@ -0,0 +1,127 @@
...
+To repeat,
+**Users and Projects can now opt out of paid ads, but will be shown Community Ads instead**.
+
+What we're doing with the money
+-------------------------------
+
+We have a couple primary things that we are spending money on,
+and plan to spend money on:
+
+* Paying full-time salaries for the primary maintainers of Read the Docs (Eric Holscher & Anthony Johnson)
```
agjohnson: Probably worth mentioning that our income is sustained by other sources, such as services and commercial hosting as well. Perhaps in more detail below though?

ads-on-other-themes.rst
```diff
@@ -0,0 +1,127 @@
...
+* Paying for contractors in order to support our users
+
+I'd like to talk a bit more each of these.
```
agjohnson: -> `a bit more on each`

ads-on-other-themes.rst
```diff
@@ -0,0 +1,127 @@
...
+Paying Salaries
+~~~~~~~~~~~~~~~
+
+Read the Docs is a massive internet service.
+We serve around 30 million _pageviews_ a month (calculated by Google Analytics, this is over 200 million _requests_ a month.)
+We are currently running the entire show with two full-time staff,
+which is incredibly understaffed compared to most internet services.
+
+We hope that over time we will be able to pay salaries for more people,
+but even if we have another $56,000 a year,
```
agjohnson: Where does 56k come from? If this isn't referenced anywhere, probably best to leave it out.

ads-on-other-themes.rst
```diff
@@ -0,0 +1,127 @@
...
+it's hard to find qualified people who will work at that rate.
+So,
+we're focusing on hiring contractors at this point.
+
+Paying to support our users
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This is the primary goal of our new income from advertising.
+We have started paying `Manuel Kauffman <https://github.com/humitos>`_ for his time helping support our users on our GitHub issue tracker.
```
agjohnson: `have started paying` implies he was working without being paid. -> `We brought on .. to help ..` perhaps

ads-on-other-themes.rst
```diff
@@ -0,0 +1,127 @@
...
+We hope to expand this program with new income,
+so that we are able to provide our open source users with better support.
+
+Please give us feedback
+-----------------------
+
+We're making these changes to hopefully continue to make a sustainable business model for our support of open source.
```
agjohnson: `to make` -> `building`

`for our support of open source` is not clear and a bit hard to parse.

PR #25, opened 2017-06-16
Add final write up of MOSS grant

ericholscher:
This is the request from MOSS:

```
Firstly, we'd like you to write an end-of-award report. This should be a
standalone document telling the story of the award, and so should contain:

* Background (previous state of code)

* What you applied to do, and how much for

* What you did, and whether it was exactly the same as what you planned

* Who worked on the project

* How long it took, and whether that was what you expected

* What positive outcomes you have seen from the work being done (uses of
the code, people appreciating new features, community growth etc.) and
what the impact has been

* What you would do differently if you did it all again

* What you plan to do next.
```

### DIFF
--- a/moss-final-report.rst
+++ b/moss-final-report.rst
@@ -7,9 +7,11 @@ MOSS Final Report
 Last year,
 we were given a :doc:`MOSS Grant<rtd-awarded-moss-grant>` to work on a few things.
 This work is live at http://www.pydoc.io/,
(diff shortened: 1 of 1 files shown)


ericholscher:
Got this to a place I'm reasonably happy, probably ready for a first pass of review.

ericholscher:
On re-read, I'm pretty happy with this post as it stands -- @agjohnson would love your feedback, especially around any large sections that might be missing, or how we might better address what Mozilla asked for.

PR #28, opened 2018-03-12
Add first draft of "Ethical Advertising Works"

ericholscher:
I'm not super happy with the results currently.
It feels like we're mixing two things:

* "It works! Thanks everyone"
* "Our vision is to take this thing that works, and expand it"

They feel very related,
it just feels like the "long list of advertisors to thank" should go _before_ the vision,
but it would be awkward structurally.

### DIFF
--- a/ethical-advertising-works.rst
+++ b/ethical-advertising-works.rst
@@ -58,14 +58,11 @@ leading by example.
 We need to have all the ads in the open source ecosystem respect user privacy,
 and then we can start to apply pressure to other parts of the tech industry.
 
-We realize that not every open source project can do ad sales,
-write advertising code,
-and all the work that we've done.
+We realize that not every open source project can do all the work that we've done.
 We want to explore how we can build an **Ethical Ad Network**,
-where user data isn't sent to a third party,
-but each project doesn't need to do the sales and marketing.
+where user data isn't sent to us,
+but each project doesn't need to do ad sales.
 
-Our approach of targeting advertising by the content of the pages allows us to show ads without knowing who a user is.
 Our ethical ad network is still in the planning phases,
 but if you know a project that might be interested in being a beta tester,
 please `get in touch`_.
--- a/ethical-advertising-works.rst
+++ b/ethical-advertising-works.rst
@@ -1,5 +1,5 @@
 .. post:: April 5, 2018
-   :tags: advertising, business, sustainability
+   :tags: advertising, business, sustainability, ethical ads
 
 Ethical Advertising Works
 =========================
@@ -30,6 +30,9 @@ We are happy to report that you can build advertising that is a win/win/win for 
 * Advertisers win because they are able to reach our audience of developers, helping to grow their business.
 * Users win because their privacy is maintained, and they continue to have access to our service.
 
+Ethical advertising on Read the Docs has not been without its challenges,
+which we will cover in future posts.
+
 Tracking users isn't required
 -----------------------------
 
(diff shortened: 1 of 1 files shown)


ethical-advertising-works.rst
```diff
@@ -0,0 +1,84 @@
...
+-----------
+
+To put it simply:
+
+    **Ethical Advertising works**.
+
+In Q1 2018,
+our estimated revenue from advertising is $75,000.
+This is enough to fund our small team of 4 people.
+We are working at below market rates for San Francisco,
```
davidfischer: I don't know if we need to call out SF in particular. None of us live in SF.

ethical-advertising-works.rst
```diff
@@ -0,0 +1,84 @@
...
+* Linode
+* Digital Impact Alliance
+* Digital Ocean
+* Exoscale
+* Nginx
+* Odoo
+* CircleCI
+* Cherry Servers
+* Uniregistry
+* OSCON
```
davidfischer: After this is a good place to ask readers if they want to advertise with us.

ethical-advertising-works.rst
```diff
@@ -0,0 +1,84 @@
...
+
+
+Next Steps
+----------
+
+We are proud to have pioneered this successful user-centric advertising model.
+
+We hope that we will act as an example for other projects that are building advertising into their business model.
+Developers have the power to shift the ad industry to be more user and privacy focused.
+We have seen other advertising platforms talk about how their ads are ethical,
```
davidfischer: I don't know specifically if others are calling out their ads as ethical but they certainly are raising advertiser standards and being slightly more transparent about what they track. Shifting this conversation at all in favor of respecting users is a big win!

ericholscher:
Would love your thoughts here too @agjohnson. I've pushed some updates that clarify the messaging, but I'm still not 100% in love with it.

ericholscher:
I feel like this could use a bit less "Eric pontificates on devs changing the world" -- I can do that on my personal blog. It could probably use more "thoughts about the industry framed from a professional perspective".

davidfischer:
If you want to talk about the ads industry, talking a bit more about the BetterAds standards and Do Not Track wouldn't be bad. Those are two small steps toward ads that respect users and see them as more than a revenue stream. We applaud companies adopting those but also think we can go further.

ethical-advertising-works.rst
```diff
@@ -0,0 +1,104 @@
...
+We can have successful advertising without tracking our users.
+So much of the trouble facing the ad industry these days is because of this massive databases of user data.
+
+We have been able to show that we can show ads without knowing exactly who our users are.
+The content they are reading is enough to show them targeted ads,
+without knowing anything about them.
+
+We're now taking a look at what we've learned,
+and how we might help more infrastructure projects reach sustainability with an ethical advertising approach.
+
```
davidfischer: I think we should be explicit in what we mean by tracking or link to a place where we have been explicit. What we mean is:

- We do not let advertisers run 3rd party ad code on Read the Docs. Ad imagery is also hosted by us.
- We do not provide advertisers with data dumps outside of aggregated reports on the clicks and views their ads got.
- We target our ads based only on geography and the details of the page being visited, not some trove of info we have on the visitor.

ethical-advertising-works.rst
```diff
@@ -0,0 +1,104 @@
...
+Towards an Ethical Ad Network
+-----------------------------
+
+**We have shown that an advertising business model that respects users can also sustain critical open source infrastructure.**
+We hope that we can act as an example for other projects that depend on ads,
```
davidfischer: I think this should be a period instead of a comma. I'm not a grammar expert though.

ericholscher:
Updated based on feedback. I think I'm ready to call this publishable. Will plan on shipping it Monday. 🚢 

I don't think it will really get much traction (sadly) because our blog has no marketing/promotion channel to users. We should really work on this some point soon.

davidfischer:
🚢 

I did notice that we had a mailchimp subscription on the blog. Does anybody actually send anything? Does it just send posts based on the RSS feed?

PR #29, opened 2018-03-29
Clean up a few bits

agjohnson:
* Drop requirements that don't work on 3.6 (not sure why werkzeug was there
  anyways)
* Update stylesheet from latest build asset from .com
* Bring fonts in

This is just a quick fix, I don't really care about style here much except for
the fact that we were using Puritan font family before and it renders poorly.
We'll address proper styles another time.


ericholscher:
Do we need a custom theme here, or should we just be using a straight alabaster theme w/ our logo? A simple example here: http://www.writethedocs.org/blog/newsletter-march-2018/

Not sure there's much value in maintaining yet another set of CSS/HTML?

PR #35, opened 2018-05-02
Login with Version Control announcement

davidfischer:
This is a pretty short blog but it should make for a good announcement. I'm aiming for a week from today which is between WTD and PyCon.

This feature was added here: https://github.com/rtfd/readthedocs.org/pull/4022.

I considered putting an image (either static or an animated gif) but I don't think it is necessary. An image of the login screen has a lot of stuff that isn't "log in with ..." related.

### DIFF
--- a/version-control-login.rst
+++ b/version-control-login.rst
@@ -1,0 +1,43 @@
+.. post:: May 9, 2018
+   :tags: feature, new-user, versioncontrol
+   :author: David
+   :location: SAN
+
+
+Version Control Log in
+======================
+
+Today we are announcing the ability to log in or sign up to Read the Docs
+with your favorite version control hosting services like GitHub, BitBucket, or GitLab.
+This was one of our most requested features and
+it has been something we've been meaning to launch for a long time.
+
+
+For new Read the Docs users
+---------------------------
+
+For new users, the sign up process is significantly streamlined.
+There's no new password to remember and when you're ready to start building your docs,
+Read the Docs will be ready with a list of your repositories to get started.
+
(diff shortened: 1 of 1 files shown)


version-control-login.rst
```diff
@@ -0,0 +1,43 @@
+.. post:: May 9, 2018
+   :tags: feature, new-user, versioncontrol
+   :author: David
+   :location: SAN
+
+
+Version Control Log in
```
humitos: I'd say that the title should reflect the "social" part of the version control. I mean, you don't log in with git, but with github.

> Social Version Control Log in

Maybe?

version-control-login.rst
```diff
@@ -0,0 +1,43 @@
...
+
+
+For new Read the Docs users
+---------------------------
+
+For new users, the sign up process is significantly streamlined.
+There's no new password to remember and when you're ready to start building your docs,
+Read the Docs will be ready with a list of your repositories to get started.
+
+Click the "sign up" button on the Read the Docs home page and then click
```
humitos: "Sign up"

version-control-login.rst
```diff
@@ -0,0 +1,43 @@
...
+the button for your favorite version control host.
```
humitos: I think it's more accurate to use the same wording you used in the first paragraph:

"version control hosting service."

davidfischer:
> My only feedback with this is that we should make sure this is out on the .com as well.

I haven't contributed on the .com but I can take a look.

agjohnson:
I opened an issue for @humitos to look at, I'd probably just vote for delaying the blog post until this is out and tested.

PR #36, opened 2018-05-04
Update about acceptable ads

davidfischer:


ericholscher:
Ah nm. Thought this was a new post. Ignore me :)

davidfischer:
The full update post is in https://github.com/rtfd/blog/pull/38. This was just a small update in case people wander across an old post.

PR #38, opened 2018-05-17
Acceptable ads and ad blocker update

davidfischer:
The goal of this post is to discuss:

- our numbers on what the acceptable ads list means for us from a revenue, views, and clicks standpoint
- a small bit of analysis on what the above says about our users (% using an ad blocker, % allow acceptable ads, etc.)
- the process for inclusion
- our inclusion on the privacy list

There isn't really a call to action in this post. It is mostly just an update.

The target audience of this post is:

- People who read the first post
- Doc authors and RTD users interested in our sustainability and open source funding

When we were first added to the EasyList, data on what that meant for other companies was hard to find. Likewise with the acceptable ads program. We had rough estimates but no idea what to expect. This shines some light on that and adds some transparency on the process for inclusion.

**Note:** whenever this blog is about to go live, I'll update the graph. I also want to put some markers on the graph showing when we were added to the easylist and when we were added to the acceptable ads list.


ad-blocker-update.rst
```diff
@@ -0,0 +1,84 @@
...
+
+People understand that advertising funds much of the web.
+As ads got larger, more annoying,
+and ad networks stored more information on millions of people, however,
+many people finally decided to install ad blockers.
+This is the `largest boycott in human history`_ and blaming users isn't the answer.
+
+Based on our experience with ad blocking and the Acceptable Ads list,
+it is clear that visitors will allow advertising that doesn't disrupt user experience
+and doesn't stalk people.
```
ericholscher: stalk is maybe a bit heavy here. It feels a bit out of place w/ the tone of the rest of the article.

I think we also need a bit more of a closing along the lines of :

We are glad to have made some progress towards regaining traffic that was lost. We are continuing on our experiment to fund open source infrastructure in a way that is a win/win for all parties, and we are glad you're following along with us.

davidfischer: I think your point about stalking is fair. I agree with the closing line. The only thing I want to avoid is calling traffic "lost". I know some people in the ad blocking world don't like that.

davidfischer:
I made the updates based on the feedback. I changed the start of the final paragraph to:

    We are glad to have made progress toward regaining our sustaining revenue.
    Read the Docs is continuing ...

Shall we aim to publish this on Monday?

ericholscher:
Works for me. 👍

davidfischer:
I'll update the image and merge it.

PR #39, opened 2018-05-25
GDPR post

davidfischer:
The goal of this post is to:

- communicate the changes (or not) we made to comply with the GDPR
- discuss advertising that complies with the GDPR because it doesn't track
- signal that we take privacy seriously

The primary audience for this post is RTD users and authors. Due to the keywords, we may get a few advertising people over the long haul but I suspect it will be small.

I didn't link to the actual [GDPR issue in the .org repo](https://github.com/rtfd/readthedocs.org/issues/3954). Do you think I should?

### DIFF
--- a/gdpr-what-it-means-for-readthedocs.rst
+++ b/gdpr-what-it-means-for-readthedocs.rst
@@ -55,7 +55,7 @@ We are taking some additional steps to protect users which we will cover in a fu
 Ethical Ads
 -----------
 
-Arguably the GDPR is a result of pernicious `privacy violations by advertising firms`_
+Arguably the GDPR is a response to pernicious `privacy violations by advertising firms`_
 so it makes sense to discuss the advertising we show on Read the Docs.
 
 While Read the Docs is an ad-supported website,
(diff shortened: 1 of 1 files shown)


gdpr-what-it-means-for-readthedocs.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Your email inbox has probably been bombarded
+over the last few days and weeks with "Updates to our Privacy Policy".
+These emails pertain to an EU law called the General Data Protection Regulation (GDPR)
+which comes into effect today.
+
+The goal of the GDPR is to put users back in control of their data.
+It is an important step toward respecting users' privacy.
+The days of collecting as much data on as many people as possible
+without consent and sharing it with anyone willing to pay for it are over.
```
ericholscher: This is a bit optimistic :)

davidfischer: Definitely. It will change some people (mostly companies with something to lose) and others will disregard it as too bureaucratic. I wouldn't mind adding a bit of humor to the post though. Perhaps something like:

    Maybe that is a bit optimistic but it has helped us build our privacy infrastructure and it does seem many companies are beginning to take privacy seriously at least if measured by the volume of privacy-related email received.

davidfischer: Eh, this is a bit repetitive with the first paragraph. I'll just publish this as-is.

PR #40, opened 2018-06-05
Do not track blog post

davidfischer:
Initially I'm targeting this post for Friday June 8. If that date slips, it should be ok.

This is a simple announcement blog post publicly discussing DNT. The target audience is RTD users. However, the call to action at the end is for people interested in the ad network.

One thing I'd like to add but don't yet have enough data on is what % of our users have DNT enabled. We will see that by a reduction in analytics pageviews. By Friday there should be enough data.

### DIFF
--- a/do-not-track.rst
+++ b/do-not-track.rst
@@ -67,8 +67,9 @@ Supporting Do Not Track is an important milestone along the way
 as it confirms our commitment to privacy in advertising.
 We believe that ads don't have to track people to be effective.
 
-Our Ethical Ad network is still in its early stages,
+Our `Ethical Ad network`_ is still in its early stages,
 but if you want to know more or you know a project that could use it, 
 please `get in touch`_.
 
+.. _Ethical Ad network: https://www.ethicalads.io/
 .. _get in touch: mailto:ads@readthedocs.org
--- a/do-not-track.rst
+++ b/do-not-track.rst
@@ -20,28 +20,39 @@ and keeping Read the Docs sustainable.
 .. _guidelines: https://www.eff.org/issues/do-not-track
 
 
+What we implemented
+-------------------
+
+Not much needed to change to make Read the Docs support DNT
+and many of these changes were already necessary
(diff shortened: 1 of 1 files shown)


do-not-track.rst
```diff
@@ -0,0 +1,63 @@
...
+Our support for Do Not Track does not change that.
+Instead, it formalizes our commitment
+to be the same as guidelines produced by the EFF and elsewhere.
+
+At Read the Docs this means:
+
+* We **do not** do behavioral ad targeting regardless of your DNT preference.
+* Our logs that store any personal data are deleted after no more than ten days.
+  We do this for all users even those who don't have DNT enabled.
+* When a user has DNT enabled, we do not send data to our analytics.
```
ericholscher: Feels like this should be above the "ads" section -- as it applies outside the ad code. Perhaps a "What we implemented" section before the "what it means for ads"?

davidfischer:
I moved the bullets and added a bit more details to fill out the section which was a bit empty without those bullets.

do-not-track.rst
```diff
@@ -0,0 +1,74 @@
...
+----------------------
+
+In a :doc:`previous post <ethical-advertising-works>`,
+we mentioned that we are taking the same developer-centric, privacy-focused
+advertising we have on Read the Docs and expanding this to a larger ad network
+for open source infrastructure.
+
+Supporting Do Not Track is an important milestone along the way
+as it confirms our commitment to privacy in advertising.
+We believe that ads don't have to track people to be effective.
```
ericholscher: Should we do a callout here for the landing page at http://www.ethicalads.io/?

davidfischer: I thought about that and I'm a little torn. It doesn't serve a *ton* of purpose yet but if we can't link to it here where would we?

Let's do it.

davidfischer:
We only have about a week's worth of data but I'm seeing about 6% fewer analytics pageviews as a result of DNT. I was thinking it would be close to 8%.

PR #42, opened 2018-07-23
API update blog

davidfischer:
This is fairly simple blog. My goals are twofold:

- Publicly commit to deprecation timelines for things that need to go away like API v1 and HTTP (not HTTPS) access. These deprecation timelines are in our docs but I don't think many people are actively looking for them. This also gives us something to point to when we remove those.
- Solicit feedback on the API and get in touch with people who are using it. This is one of the big problems with no authentication on the API (v1 or v2). We have no real idea who is using it.

I am on the fence about mentioning API v3. I think that changes the tone of the blog and it's odd to be talking about how you just documented API v2 when you're about to replace it.

### DIFF
--- a/api-v2.rst
+++ b/api-v2.rst
@@ -1,0 +1,44 @@
+.. post:: July 26, 2018
+   :tags: api, feature, reference
+   :author: David
+   :location: SAN
+
+Read the Docs Public API
+========================
+
+Recently, we revamped Read the Docs' public API.
+Previously, our latest API (v2) was used by our build processes
+but not heavily used by outside users.
+
+As part of this process, we put effort into making sure the API is easy to use
+to access Read the Docs *projects*, *builds*, and *versions*
+and that the `documentation`_ is up-to-date.
+
+.. _documentation: https://docs.readthedocs.io/en/latest/api/v2.html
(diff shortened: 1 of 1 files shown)


api-v2.rst
```diff
@@ -0,0 +1,44 @@
...
+Read the Docs Public API
+========================
+
+Recently, we revamped Read the Docs' public API.
+Previously, our latest API (v2) was used by our build processes
+but not heavily used by outside users.
+
+As part of this process, we put effort into making sure the API is easy to use
+to access Read the Docs *projects*, *builds*, and *versions*
+and that the `documentation`_ is up-to-date.
```
humitos: Doesn't worth to mention what are those changes? Like adding the `project__slug` filter and `active`, for example?

davidfischer: I added a note about filtering specifically. Considering there wasn't any documentation before, I think mentioning `project__slug` specifically isn't necessary. Do you think that's ok?

humitos: The note is good.

I'd say that it's not _necessary_. It would depend on how technical we want to be in the post. It may add more noise than clarification, though.

PR #45, opened 2018-08-03
HTTPS for custom domains blog post

davidfischer:
There are a few goals:

1. Credit Cloudflare for all their help. The SSL for SaaS package is not cheap and we'd still be building comparable functionality if they didn't help with it. I also want to delve a little bit into the value they provide without being too sales-ey. If I'm too sales-ey, let me know.
1. Get the word out! At PyCon this year, this was the most often talked about feature I heard from our users.
1. Let users who have an old setup know that they can take action to enable HTTPS for their domain.

The target audience for this post is maintainers of Read the Docs projects.

### DIFF
--- a/https-for-custom-domains.rst
+++ b/https-for-custom-domains.rst
@@ -1,0 +1,75 @@
+.. post:: August 10, 2018
+   :tags: security, feature
+   :author: David
+   :location: SAN
+
+
+HTTPS for Custom Domains
+========================
+
+Read the Docs hosts documentation for over 80,000 open source projects
+and over 2,500 of those projects are hosted on their own individual domains.
+Documentation hosted on ``*.readthedocs.io`` has supported HTTPS for a number of years,
+but one of our most requested features was to make HTTPS on other domains easy.
+Today we are happy to announce that **Read the Docs supports HTTPS on custom domains**!
+
+Earlier this year, Cloudflare contacted us to support HTTPS
+for the thousands of open source documentation projects on their own domains.
+They generously provided us with their `SSL for SaaS`_ package
+to ease the integration on our side.
+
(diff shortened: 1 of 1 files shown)


https-for-custom-domains.rst
```diff
@@ -0,0 +1,76 @@
...
+Configuring your domain
+-----------------------
+
+If you followed our `documentation for custom domains`_
+and you are using a ``CNAME`` to point to Read the Docs,
+no action is required on your part.
+
+If you have had your Read the Docs setup for many years
+you may have followed some older directions.
+Depending on your setup, we may not have been able to issue a certificate for your domain.
```
ericholscher: This is a big ambiguous. We should be explicit here with what a "wrong" config looks like.

https-for-custom-domains.rst
```diff
@@ -0,0 +1,76 @@
...
+
+We identified a number of challenges including:
+
+- Retrying failed requests to issue certificates
+- Handling when users remove DNS records that would allow us to issue certificates
+- Getting our web servers and load balancers to handle dynamically adding and updating certificates on the fly
+- Handling recurring tasks to re-issue updated certificates before they expire
+
+Cloudflare reduced most of this complexity to a few API calls made when a new domain is configured or removed.
+Without their support, we would almost certainly still be working on this feature.
```
ericholscher: Kill the `almost certainly`.

davidfischer:
I made the changes based on feedback.

I think debugging custom domains should live [in the docs](https://docs.readthedocs.io/en/latest/alternate_domains.html). I forgot to mention that there is a pending update to those docs which should go live before this goes live: https://github.com/rtfd/readthedocs.org/pull/4425

ericholscher:
💯 -- Good use of a time machine ;)

PR #46, opened 2018-08-06
Planned Azure migration

davidfischer:
My goals with this post are to:

- Inform users about timeframe of site maintenance
- Inform users about what is down (and what shouldn't be)
- Have something we can point to from Twitter, etc.

Should we give some more credit to Microsoft? Should we talk a bit about some of the features we are excited about on Azure?

### DIFF
--- a/azure-migration.rst
+++ b/azure-migration.rst
@@ -9,7 +9,6 @@ Planned Move to Azure
 
 We wanted to make you aware that on Saturday August 18 at 10:00am PDT (1:00pm EDT, 17:00 UTC),
 Read the Docs will be having a **scheduled downtime of approximately 4 hours**.
-During this time, we will be moving our service to our new home on Microsoft Azure.
 
 To ensure the stability and performance of our system,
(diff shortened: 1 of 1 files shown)


azure-migration.rst
```diff
@@ -0,0 +1,22 @@
...
+   :tags: azure, downtime, infrastructure
+   :author: David
+   :location: SAN
+
+
+Planned Move to Azure
+=====================
+
+We wanted to make you aware that on Saturday August 18 at 10:00am PDT (1:00pm EDT, 17:00 UTC),
+Read the Docs will be having a scheduled downtime of approximately 2 hours.
```
ericholscher: We should give ourselves more buffer than this. If we think it should only take 2 hrs best case, we should schedule the window for at least 4, because something will go wrong, and if we end early, people are happy.

davidfischer:
I increased the maintenance window to 4 hours and credited Microsoft while mentioning some cool features.

PR #50, opened 2018-10-29
First blog on ads for hiring

davidfischer:
This is a post around hiring on Read the Docs with advertising. The target of this blog post is people/companies who are recruiting. It is a little marketing focused but my plan is to run a few house ads that point to this post and at least 1 twitter post promoting it.

My goal is to get the word out that RTD can be used for hiring. I don't know if this will immediately result in people contacting us but it also just puts the idea into people's heads that they can hire with RTD and it puts something out there that can be found by search engines and whatnot down the line.

## Thoughts/Questions

* Should we have numbers on the graph? It really shows that our CTR is really low although there are a number of reasons for that (ads aren't always above the fold being the big one). We could remove it but then the image is a little bare.
* I didn't call out any specific advertisers here and I'm not sure I should. It would be nice to get a few testimonials around hiring though.


davidfischer:
Maybe mention something about how the ad helps the developer ecosystem (by helping RTD) while also helping businesses hire people?

ericholscher:
> Should we have numbers on the graph? It really shows that our CTR is really low although there are a number of reasons for that (ads aren't always above the fold being the big one). We could remove it but then the image is a little bare.

I think we keep the numbers. No real point in hiding the truth, I think. 

> I didn't call out any specific advertisers here and I'm not sure I should. It would be nice to get a few testimonials around hiring though.

This would be useful for a landing page for sure.

hiring-developers-with-readthedocs.rst
```diff
@@ -0,0 +1,79 @@
...
+Developers are not always actively job seeking
+by browsing job boards or company careers pages.
+However, they are on Read the Docs reading about the libraries and frameworks they use.
+Even when people aren't actively looking for a new job,
+many are open to exploring new opportunities.
+We found that developers engage with hiring ads at a higher rate than any other kind of advertising.
+
+.. figure:: img/2018-ctr-adtype.png
+   :alt: Click-through rate by ad type Q3 2018
+   :width: 100%
```
ericholscher: `Other` here is a bit funny, as it's among the highest ranking and isn't clear what that means. I'd maybe remove it?

ericholscher: Also it's probably useful to sort them by % or something to make it a bit easier to process.

davidfischer:
I updated the graphic and built out the narrative on why hiring ads work a bit. If you are happy with this post, I'll push it live tomorrow.

hiring-developers-with-readthedocs.rst
```diff
@@ -0,0 +1,83 @@
...
+
+
+Hiring with Read the Docs works 
+-------------------------------
+
+Developers are not always actively job seeking
+by browsing job boards or company careers pages.
+However, they are on Read the Docs reading about the libraries and frameworks they use.
+Even when people aren't actively looking for a new job,
+many are open to exploring new opportunities.
```
humitos: not my field but shouldn't be "to explore" here?

hiring-developers-with-readthedocs.rst
```diff
@@ -0,0 +1,83 @@
...
+
+On Read the Docs, we found that developers engage with hiring ads
+at a higher rate than any other kind of advertising.
```
humitos: If you were a company that is advertising at RTD with ads that it's about your product, how would you read this sentence? Would it "hurt" you?

humitos: Anyway, the graphic says exactly the same without words, though.

hiring-developers-with-readthedocs.rst
humitos: is it possible to sort increasing instead of decreasing? just by looking at it (without reading the numbers and the meaning) seems like things are going down :)

It's more a psychological thing, I guess.

hiring-developers-with-readthedocs.rst
```diff
@@ -0,0 +1,83 @@
...
+
+
+Be specific and find the right developers
+-----------------------------------------
+
+If you are looking for Python developers,
+you can configure your hiring ad to only show on Python related documentation.
+You can further refine your audience to Read the Docs visitors in specific regions.
+Whether you are looking for a remote Python data scientist or a blockchain developer in Switzerland,
+be as specific as possible.
```
humitos: As far as I know, we don't have too many filters to "be as specific as possible".

hiring-developers-with-readthedocs.rst
```diff
@@ -0,0 +1,83 @@
...
+Set the right expectations
+--------------------------
+
+In any type of advertising, setting appropriate expectations can make the difference
+between a lackluster and successful campaign. In hiring, it is even more critical.
+This means making sure that your hiring ad for a front-end JavaScript developer
+links directly to your JavaScript developer job listing page where people can click "apply".
+Don't link to a listing of all of your open positions unless your hiring ad sets that expectation.
+Nothing is worse for the developer than clicking on a hiring ad
+and not knowing what to do next.
```
humitos: Excellent paragraph!

hiring-developers-with-readthedocs.rst
davidfischer: I don't believe there is a grammatical issue with the last sentence ending "...  many are open to exploring new opportunities."

hiring-developers-with-readthedocs.rst
davidfischer: > Would it "hurt" you?

Well it's a double edged sword. For cpc ads it will cost more (we do cpm as well where it wouldn't). However the whole point of ads is for people to engage with them.

hiring-developers-with-readthedocs.rst
davidfischer: Well, it was more in line with be specific about the job description. You are right that we don't have fantastic keyword targeting for jobs yet. However, that's something that can be solved over time.

PR #55, opened 2019-01-03
Initial swing at the 2018 stats post.

ericholscher:
Revenue numbers still aren't complete,
just a ballpark

### DIFF
--- a/read-the-docs-2018-stats.rst
+++ b/read-the-docs-2018-stats.rst
@@ -4,7 +4,7 @@
 Read the Docs 2018 Stats
 ========================
 
-2017 was another good year for Read the Docs.
+2018 was another good year for Read the Docs.
 We've settled into a sustainability model that is working for us,
 and have a team of 5 folks working full-time on the project.
 
--- a/read-the-docs-2018-stats.rst
+++ b/read-the-docs-2018-stats.rst
@@ -91,6 +91,10 @@ We are also exploring sharing revenue with our documentation authors.
 There are currently 3 projects enrolled in a revenue share,
 and we're hoping to expand that in 2019 to better support open source projects.
 
+
+The project is now on firm finaicial footing.
+We have a cash reserve that will allow us to pay all of our staff for 3 months in the event of a loss of revenue,
+which allows us to be more aggressive with our revenue going forward. 
 This funding allowed us to expand from 4 to 5 full-time folks working on the project.
 
 We also moved hosting to Azure_ this year in August from Rackspace_,
--- a/read-the-docs-2018-stats.rst
+++ b/read-the-docs-2018-stats.rst
@@ -31,8 +31,8 @@ Page Views
 
 Our stats:
 
-* 400 Million Page Views (up from 338M)
-* 77 Million Unique Visitors (up from 75M)
+* 400 Million Page Views (up from 338M, 18%)
+* 77 Million Unique Visitors (up from 75M, 3%)
 
 .. From Google Analytics
 
@@ -43,8 +43,8 @@ Site Stats
 
 The stats, in total numbers:
 
-* 100,000 projects (up from 77k)
-* 155,000 users (up from 103k)
+* 100,000 projects (up from 77k, 30%)
+* 155,000 users (up from 103k, 50%)
 
 We have been battling spam quite heavily again this year,
 so these numbers might be a bit skewed.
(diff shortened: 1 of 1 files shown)


ericholscher:
@rtfd/core Curious if there are any other stats that you would find interesting to publish & track. This is mostly what we've been doing in the past, but always happy to publish new stats if they are useful to folks.

read-the-docs-2018-stats.rst
```diff
@@ -0,0 +1,107 @@
...
+Our stats:
+
+* 400 Million Page Views (up from 338M)
+* 77 Million Unique Visitors (up from 75M)
+
+.. From Google Analytics
+
+These numbers are not as accurate this year as in the past,
+because we have :doc:`implemented Do Not Track support </do-not-track>` to our Google Analytics setup.
+So we're losing a non-trivial number of users in that
```
stsewd: A missing period at the end of this sentence

davidfischer: I would actually get a little more specific and say something to the affect of "since we implemented Do Not Track in May, our analytics pageview figures understate our actual traffic".

read-the-docs-2018-stats.rst
```diff
@@ -0,0 +1,107 @@
...
+
+.. _people: https://github.com/rtfd/readthedocs.org/graphs/contributors?from=2018-01-01&to=2018-12-31&type=c
+.. _commits: https://github.com/rtfd/readthedocs.org/commits/master
+.. _issues: https://github.com/rtfd/readthedocs.org/issues?utf8=%E2%9C%93&q=is%3Aissue++created%3A2018-01-01..2019-01-01+
+
+Funding
+-------
+
+* $400,000 in revenue (up from $250,000)
+    * $300,000 from advertising
```
davidfischer: Do you think it's worth mentioning the impact that ad blockers had on our ad revenue over the year? While it's a bit fuzzy, I think we could put together a pretty good estimate.

ericholscher: Yea, could be interesting. Definitely worth mentioning and linking to, at least. But if we have approx. numbers that would be 👍

ericholscher: I updated it with percentages. That seems probably good enough, but we could also just do `(Q3 + Q4 ad revenue) * .15` to get a number probably.

davidfischer: 15% is pretty much right on the money [based on our stats](https://docs.readthedocs.io/en/latest/advertising/ad-blocking.html#statistics-and-data).

The only thing I'd say is to not use the word "lost" when talking about ads being blocked. Some folks especially in the ad blocker communities don't like that. I'd phrase it more along the lines of "For the second half of the year, our ad revenue was down about 15%".

read-the-docs-2018-stats.rst
```diff
@@ -0,0 +1,110 @@
...
+So for the second half of the year we lost around 15% of revenue from advertising.
+
+This funding allowed us to expand from 4 to 5 full-time folks working on the project.
+
+We also moved hosting to Azure_ this year in August from Rackspace_,
+who both have sponsored our hosting.
+
+.. _Rackspace: http://rackspace.com/
+.. _Azure: https://azure.microsoft.com/en-us/
+.. _Ethical Advertising: http://docs.readthedocs.io/en/latest/ethical-advertising.html
```
humitos: This link is not used in the post.

read-the-docs-2018-stats.rst
```diff
@@ -0,0 +1,110 @@
...
+
+This year, we had:
+
+* 28 `people`_ who committed code to the main repository (up from 23)
+* 3774 `commits`_ (up from 1058)
+* 797 `issues`_ - 189 open, 608 closed (up from 513)
+
+You can see the number of commits and issues has gone up quite a bit.
+This is largely because of the additional staffing we have had on the project,
+allowing us to be more active to both our users and to release more features.
```
agjohnson: I like the idea of putting these into percentage change basis as well, just to describe some of these figures in an easier way. Commits increasing 256% is pretty incredible.

read-the-docs-2018-stats.rst
```diff
@@ -0,0 +1,110 @@
...
+.. _commits: https://github.com/rtfd/readthedocs.org/commits/master
+.. _issues: https://github.com/rtfd/readthedocs.org/issues?utf8=%E2%9C%93&q=is%3Aissue++created%3A2018-01-01..2019-01-01+
+
+Funding
+-------
+
+* $294,000 in revenue from readthedocs.org (up from $250,000)
+    * $278,000 from advertising (up 59%)
+    * $16,000 from Gold users (up 16%)
+    * We have additional revenue from our commercial offering at readthedocs.com, but aren't including that currently.
```
agjohnson: `aren't including that currently` to something like `aren't including that in our community funding overview`.

Or some similar way of communicating we consider the revenue separate from community funding

PR #56, opened 2019-01-12
Add blog post on incoming webhook deprecations

agjohnson:
Addresses GitHub and our other legacy webhooks. Does address some of the other
changes to RTD, those will be separate posts.

### DIFF
--- a/webhook-deprecations.rst
+++ b/webhook-deprecations.rst
@@ -1,4 +1,4 @@
-.. post:: Jan 11, 2019
+.. post:: Jan 21, 2019
     :tags: deprecation
     :author: Anthony
 
--- a/webhook-deprecations.rst
+++ b/webhook-deprecations.rst
@@ -6,12 +6,20 @@ Incoming Webhook Deprecations
 =============================
 
 In the coming weeks and months, Read the Docs will be moving some projects away
-from our legacy incoming webhooks. About 2 years ago, these incoming webhooks
-were replaced with project webhook integrations, at which point we improved
-security and added several features to these incoming webhooks. New projects
-have been using these integrations by default since, but it's possible that
-your project is still using these deprecated incoming webhooks if your project
-was configured on Read the Docs more than 2 years ago.
+from our legacy incoming webhooks, towards our per-project webhook integrations.
+
+Our legacy incoming webhooks were our first attempt at allowing providers like
+GitHub to automatically trigger builds on for projects on Read the Docs. These
+webhooks lacked a number of security features, and so, about two years ago, we
+replaced these with per-project webhook integrations instead. We added a number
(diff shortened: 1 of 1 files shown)


webhook-deprecations.rst
```diff
@@ -0,0 +1,41 @@
+.. post:: Jan 11, 2019
+    :tags: deprecation
+    :author: Anthony
+
+Incoming Webhook Deprecations
+=============================
+
+In the coming weeks and months, Read the Docs will be moving some projects away
+from our legacy incoming webhooks. About 2 years ago, these incoming webhooks
+were replaced with project webhook integrations, at which point we improved
```
davidfischer: I think we need some explanation of the difference between "legacy webhooks" and "project webhooks". I personally am not super familiar with the difference in our code and I was a bit confused by these couple sentences.

webhook-deprecations.rst
```diff
@@ -0,0 +1,41 @@
...
+security and added several features to these incoming webhooks. New projects
+have been using these integrations by default since, but it's possible that
+your project is still using these deprecated incoming webhooks if your project
+was configured on Read the Docs more than 2 years ago.
+
+In order to continue building automatically, projects should reconfigure their
+repository to use a new project webhook integration instead. For more
+information, see our docs on :ref:`readthedocs:webhook-creation`.
```
davidfischer: I was going to say that we should probably just have the basics of the docs inline here but the process is somewhat complicated. Are you worried about a significant dropoff in projects?

webhook-deprecations.rst
humitos: Agreed.

From a technical point of view, "legacy webhooks" are the URLs

* /github
* /gilab
* /build
* /bitbucket

and the new projects integrations are:

* /webhook/github
* /webhook/*

agjohnson: Yeah, hard to explain at a high level. The new webhooks are attached to projects, the old webhooks were public -- so you could trigger a build for whatever project you wanted. I debate getting into a deep discussion in the blog post, but probably does warrant something in the docs maybe?

webhook-deprecations.rst
agjohnson: I think we should be able to make sure most projects are configured. We'll no more on numbers once we wait a week and see who has an integration.

webhook-deprecations.rst
agjohnson: Okay, i tried to clarify. Feel free to merge if it makes more sense now. It's rather hard to distinguish the two, we never really talked about or exposed the legacy webhooks. They were hidden and only internal.

agjohnson:
I'm going to merge this, if there are any changes, let's sneak them in tomorrow.

PR #57, opened 2019-01-12
Post on defaulting to Python 3

agjohnson:
We're not yet ready for this to be merged, we have some final pieces and
a release of our docker image to perform before going official with
this.

### DIFF
--- a/default-python-3.rst
+++ b/default-python-3.rst
@@ -1,15 +1,27 @@
-.. post:: Jan 11th, 2019
+.. post:: Feb 8th, 2019
    :author: Anthony
    :tags: python
 
-Defaulting to Python 3
-======================
+Defaulting New Projects to Python 3
+===================================
 
-Starting February 1st, 2019, new projects getting started with Read the Docs
-will use Python 3 by default. While it is stil possible to configure your
-project to use Python 2.7 with :ref:`our config file <readthedocs:yaml__python>`,
-we think it's important for the ecosystem to help push projects towards adopting
-Python 3.
+New projects that are just getting started with Read the Docs will now use
+Python 3 by default. While it is still possible to configure your project to use
+Python 2.7 with :doc:`our configuration file <readthedocs:config-file/v2>`, we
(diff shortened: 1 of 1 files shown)


default-python-3.rst
```diff
@@ -0,0 +1,14 @@
+.. post:: Jan 11th, 2019
+   :author: Anthony
+
```
davidfischer:
```suggestion
   :tags: python

```

By adding the python tag, this ensures that the post gets put into Planet Python's RSS feed.

default-python-3.rst
```diff
@@ -0,0 +1,14 @@
...
+======================
+
+Starting February 1st, 2019, new projects getting started with Read the Docs
+will use Python 3 by default. While it is stil possible to configure your
+project to use Python 2.7 with :ref:`our config file <readthedocs:yaml__python>`,
+we think it's important for the ecosystem to help push projects towards adopting
+Python 3.
+
+For the time being, Python 3.6 will be our default Python 3 version, however we
+will have Python 3.7 available for testing very shortly as well.
```
humitos: I wouldn't say "for testing" here.

default-python-3.rst
humitos: I like the idea of "going global" with this post.

humitos:
Linking the PR that contains the docker image modifications for this https://github.com/rtfd/readthedocs-docker-images/pull/84 and the issue onder .org repo defaulting to Python3 https://github.com/rtfd/readthedocs.org/pull/3581

humitos:
Unblocking this since we are close to make Python3 the default. We should take a look at the feedback only.

default-python-3.rst
agjohnson: We're defaulting to 3.7 now anyways :) i'll update

agjohnson:
I updated this post a bit. In intersphinx links, I noticed our RTD docs have a bad version number, so this might depend on rtfd/readthedocs.org#5259 as well.

agjohnson:
The buttons have been pushed.

PR #67, opened 2019-02-13
Replace "Read The Docs" by "Read the Docs" in all posts

humitos:
Just a branding fix.


ericholscher:
We run `vale` on the WTD repo to make sure this doesn't happeN: https://github.com/writethedocs/www/blob/master/vale/WTD/Branding.yml

PR #68, opened 2019-05-22
Ethical ads update

davidfischer:
This post contains:

- A financial update. We last gave a [partial update in January](https://blog.readthedocs.com/read-the-docs-2018-stats/) and our last full update was [a year ago](https://blog.readthedocs.com/ethical-advertising-works/).
- A progress update on changes over the last year. This is mostly recap.
- ~~New ad format rollout and timeline. I do want to make sure I'm not burying the lede here but I also want to explain what we use money for and how we help the community before telling people with custom themes that they will get ads.~~

One thing I would like to get down a little better is our exact upsell path for companies who don't want ads. I feel like we are close there but I think we could use a little refinement.


ethical-advertising-update-2019.rst
```diff
@@ -0,0 +1,117 @@
...
+
+Despite a slow start to the year, we expect to earn about $75,000 from advertising in Q2.
+This brings our advertising revenue back to the level
+before our ads were :doc:`added to major ad blocking lists <ads-and-adblocking>`.
+
+While we do earn money from other sources such as our commercial offering,
+advertising continues to be our largest source of revenue and has allowed us
+to expand our team to five full-time people.
+Compared with three years ago when we were scraping along with very little money,
+our community issue tracker and support requests receive prompt responses.
```
humitos: :100:

ethical-advertising-update-2019.rst
```diff
@@ -0,0 +1,117 @@
...
+We purposely did this slowly to make sure our ads look integrated with Read the Docs and less obtrusive to users.
+**This new ad format will allow us to show advertising on all documentation themes.**
+
+We plan to roll this new ad format out more widely in stages:
+
+* To start with, we are going to enable this ad format for community and house advertising
+  on pages that already have an ad on them.
+  As always, we never display two ads on the same page so this ad will be instead of any other ad.
+* After that, we will allow paid advertisers to use this new format.
+* We will email users of custom documentation themes to let them know this change is coming.
```
humitos: How do we know that the docs is using a different theme?

ethical-advertising-update-2019.rst
```diff
@@ -0,0 +1,117 @@
...
+* At first, we will show only community and house ads on these custom themes.
+* Finally, we'll turn on our normal ad rotation to all themes. We expect to complete this by July.
+
+We will closely monitor feedback from document authors, readers, and advertisers during this process.
+If you have questions or comments on this, please `let us know`_.
+
+.. _let us know: mailto:ads@readthedocs.org?subject=New+Ad+Format+Feedback
+
+Users can opt-out of paid advertising themselves and for any projects they maintain
```
humitos: I think with the latest changes it's not possible to remove ads for a project anymore.

ethical-advertising-update-2019.rst
davidfischer: We can get it off of disk or from analytics.

ethical-advertising-update-2019.rst
davidfischer: They can't remove all ads without speaking to us but they can remove paid ads. For example, we could remove paid ads from the "docs" project here: https://readthedocs.org/dashboard/docs/advertising/

ethical-advertising-update-2019.rst
```diff
@@ -0,0 +1,117 @@
...
+Results
+-------
+
+**Our ethical advertising model is still going strong.**
+
+Despite a slow start to the year, we expect to earn about $75,000 from advertising in Q2.
+This brings our advertising revenue back to the level
+before our ads were :doc:`added to major ad blocking lists <ads-and-adblocking>`.
+
+While we do earn money from other sources such as our commercial offering,
```
ericholscher:
```suggestion
While we do earn money from other sources such as our `commercial offering <https://readthedocs.com>`_,
```

ethical-advertising-update-2019.rst
```diff
@@ -0,0 +1,117 @@
+.. post:: May 28, 2018
+   :tags: advertising, business, sustainability, ethical ads
+   :author: David
+   :location: SAN
+
+Ethical Advertising Update
```
ericholscher: Needs a better title. I think we're really doing two things here, which maybe should be split into two posts:

* An ethical ads update
* Announcing a new ad format and expanding where we show ads

It definitely feels like burying the lede to do the later in a blog post called "ad updates"

davidfischer: Agreed 100% on the title.

I completely understand this concern although I do want to build value for what we're doing before I tell people more ads are coming.

I see a few options here:

- The title could be changed to something like "Ethical Advertising Update & Coming Changes" to make sure to note that changes are coming.
- We could break the blog into two although then we are going to have a much shorter "build value" section before we talk about more ads. I also feel that the [blog for Alabaster ads](https://blog.readthedocs.com/ads-on-other-themes/) was a little awkward in that it just informed people ads were coming without doing enough to explain how users benefit from additional support, etc. Or rather, it reversed what I think of as the logical order of those things.
- We could refactor the blog a bit to mention some of the pending changes earlier and go into more details further down.

Thoughts?

ericholscher: I think we need to at least mention a new ad format in the title and in the introduction. I don't think it's awkward to tell people ads are coming, when we want to communicate to them ads are coming :) If we're going to email people this post who are going to get ads on their docs to explain this to them, the current structure is not ideal.

davidfischer: Sounds good. I'll change this post to just be an ethical ads update. I'll create another post for the changes.

davidfischer:
I have updated this post to be purely an advertising update. I tease the new ad format but we'll detail it in a different post.

This post does feel a bit status-ey. While that is probably the main goal of the post, it might be a bit more interesting if it was a bit more forward looking.

While the title of the post "Ethical Advertising Update" is more accurate now, it still feels a bit weak. How about:

- Ethical Advertising at Read the Docs: A Year in Retrospect
- Ad Funding at Read the Docs and What's Next for Ethical Advertising

ericholscher:
Of the two, I like the second better. Agreed that an "update" isn't super exciting -- we could also editorialize a bit and say something like "Ethical Advertising Update: Continuing to make money without tracking our users" or something cheeky like that. That isn't it, but something along those lines could also work.

davidfischer:
Updated the title with a bit of editorializing. I think this is just about ready to go.

PR #71, opened 2019-07-09
Add a blog post for custom CSS and JS

davidfischer:
Most of this content comes straight from https://github.com/readthedocs/readthedocs.org/pull/5893.

I modified the `line-height` of the blog which was set to `1` by default which makes code snippets so tight that they overlap a bit.

### DIFF
--- a/custom-css-and-js-in-sphinx.rst
+++ b/custom-css-and-js-in-sphinx.rst
@@ -30,7 +30,9 @@ Adding additional stylesheets or scripts
 
 If your custom stylesheet is ``_static/css/custom.css``,
 you can add that CSS file to the documentation using the
-Sphinx option `html_css_files`_::
+Sphinx option `html_css_files`_:
+
+.. code-block:: python
 
     ## conf.py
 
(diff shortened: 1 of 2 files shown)


custom-css-and-js-in-sphinx.rst
```diff
@@ -0,0 +1,94 @@
...
+
+
+Adding additional stylesheets or scripts
+----------------------------------------
+
+If your custom stylesheet is ``_static/css/custom.css``,
+you can add that CSS file to the documentation using the
+Sphinx option `html_css_files`_::
+
+    ## conf.py
```
humitos: nit: it would be good to use `.. code-block:: python` directive so we get the highlight here and in the following one as well.

davidfischer: The blog highlights Python by default. It's better to be explicit though so I'll update it.

davidfischer:
> is there an SEO benefit or loss to having this content in multiple places, or is this mostly just to highlight it for human users?

If it was virtually entirely duplicate content, it would be negative for SEO. Since it isn't and the two posts link to each other and build on each other, it should be positive.

I intend to roll this out tomorrow.

PR #74, opened 2019-07-23
New blog post: Documentation Versioning

### LINKED ISSUE #53: "How to do documentation versioning" blog post
It would be good to have a blog post explaining "How to do documentation versioning" and the different approaches that you can do.

There are some articles already,

* https://docs.readthedocs.io/en/latest/versions.html#how-we-envision-versions-working
* http://ericholscher.com/blog/2016/jul/1/sphinx-and-rtd-for-writers/#recommended-versioning-system
 
The problem that I feel with tags as documentation versions is that you are tied or "blocked" there and if you find a simple typo or a code example that doesn't work you can fix it anymore because the tag is frozen and that doesn't worth a new release of the whole software.

Another example is that you can't change the theme or any Sphinx setting or Read the Docs YAML config option at all.

Mentioning this potential issues in the future could make the user to think twice what approach he/she want to follow when creating RTD versions for the documentation.

humitos:
This is my first post on Read the Docs' blog.

I tried to achieve what we have talked about Documentation Versioning mentioning what are the most common problems and what's the solution that we are proposing.

Overall, I like the post, but I feel that the solution proposed needs more strength to catch people's attention.

I'm open to suggestion on every single word --or even, rewriting it completely ;)

Rendered version: https://readthedocs-blog--74.org.readthedocs.build/documentation-versioning/

Closes #53


documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+This may not seems like a problem at first sight since that's the main concept of :abbr:`VCS` tags.
+Although, when talking about documentation versions,
+sometimes happen that after making a release there is a typo in the docs, or a small code example that does not work,
+or a sentence that is confusing, etc. and we want to fix it to avoid this inconveniences.
+
+Making a new tag and release just for these small changes does not worth the effort but also it may confuse our users.
+In other words, having a ``1.0.4`` release that only fixes a sentence in the documentation is not optimal.
+
+Another problem that we have found with this approach is when new features are released on Read the Docs.
+Applying this feature to all of our documentation versions --new ones and old ones, it's a problem.
```
stsewd: mdash

```suggestion
Applying this feature to all of our documentation versions —new ones and old ones, it's a problem.
```

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+   :tags: vcs, versioning
+   :author: Manuel
+   :location: BCN
+
+.. meta::
+   :description lang=en:
+
+      Recommendations about how to do documentation versioning.
+
+Documentation Versioning: Good Practices
```
ericholscher:
```suggestion
Documentation Versioning: Best Practices
```

Believe this is probably the term of art.

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+========================================
+
+Project releases and documentation versions are strongly related each other.
+We usually want to have a release ``1.0.3`` with it's own documentation version ``1.0.3`` that's specific for that release.
```
ericholscher:
```suggestion
We usually want to have a release ``1.0.3`` with it's own documentation version ``1.0.3`` for that specific release.
```

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+
+In practice, when we make a new release of our software,
```
ericholscher: Less is more in writing :)

```suggestion
When we make a new release of our software,
```

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+we create a new tag in our :abbr:`VCS` for this release and activate this new version on Read the Docs.
+At this point, Read the Docs will build and publish the new documentation version ``1.0.3`` for us.
+
+This approach of "VCS tag" mapped to a "Read the Docs version" works in most of cases.
```
ericholscher:
```suggestion
This approach of "VCS tag mapped to a Read the Docs version" works in most cases.
```

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+Although, there could be some problems that may arise in time and that we can prevent using a different strategy.
```
ericholscher:
```suggestion
There are sometimes problems with this approach that we can prevent using a different strategy.
```

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+
+
+Problems of using tags for documentation versions
+-------------------------------------------------
+
+The main problem of using VCS tags to build documentation versions is that they are static.
+This means, that once the tag is created it will never be modified.
```
ericholscher:
```suggestion
Once a tag is created it can never be modified.
```

Could use a footnote saying "it's technically possible, but not recommended by all packaging tools"

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+
+This may not seems like a problem at first sight since that's the main concept of :abbr:`VCS` tags.
```
ericholscher:
```suggestion
This may not seem like a problem at first, since that's the main concept of :abbr:`VCS` tags.
```

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+Although, when talking about documentation versions,
```
ericholscher:
```suggestion
When talking about documentation versions,
```

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+sometimes happen that after making a release there is a typo in the docs, or a small code example that does not work,
```
ericholscher:
```suggestion
sometimes after making a release there is a typo in the docs, or a small code example that doesn't work,
```

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+or a sentence that is confusing, etc. and we want to fix it to avoid this inconveniences.
```
ericholscher:
```suggestion
and we want to fix it.
```

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Making a new tag and release just for these small changes does not worth the effort but also it may confuse our users.
```
ericholscher:
```suggestion
Making a new tag just for these small changes isn't worth the effort, and  also it can confuse users.
```

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+In other words, having a ``1.0.4`` release that only fixes a sentence in the documentation is not optimal.
```
ericholscher:
```suggestion
```

documentation-versioning.rst
ericholscher:
```suggestion
Adding this new feature to all of our documentation versions is impossible. 
```

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+Similarly, if we want to install a new Sphinx extension in all our versions,
+update one of its settings, change the theme,
+or even change a build config in the `.readthedocs.yml`_,
+we will be "blocked" by the tags being static.
+
+
+Recommended versioning strategy
+-------------------------------
+
+The workflow that we recommend to versioning your documentation is by having release branches instead.
```
ericholscher:
```suggestion
**The workflow we recommend for versioning your documentation is having release branches.**
```

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+With this approach, if a reader detects a typo in the documentation,
+we can just fix the typo and commit and push to this ``1.0.x`` release branch.
+By doing this, the documentation will be automatically updated and published on Read the Docs.
+
+Following our example, this strategy would be:
+
+* ``1.0.x`` is a release branch for the ``1.0`` version
+* ``1.0.3`` is a similar tag for your ``1.0.x`` branch, with the latest release
+* ``1.0.0`` is the first tag for your ``1.0.x`` branch.
+
```
ericholscher: I feel like we need to expand more on this. This mostly matches what I had here: https://www.ericholscher.com/blog/2016/jul/1/sphinx-and-rtd-for-writers/#recommended-versioning-system -- but what do we do with the tags and old links to `1.0.3` when `1.0.4` is released? I think we need a bit more guidance there.

humitos: Yes. I took the inspiration from there and my goal was to expend it and make it better. I failed :slightly_smiling_face:

humitos:
> I think it needs a bit more of a workflow defined.

I agree. I'm working on trying to make it better.

> I think GitHub Flow is perhaps a good model to follow here.

I think that explaining this flow is out of the scope of this blog post. Although, I think it would be good to link this page, which is really nice: https://guides.github.com/introduction/flow/. It doesn't talk about "release branches", though.

humitos:
You may probably refers to `git-flow` instead of GitHub Flow ;)

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+Documentation Versioning: Best Practices
+========================================
+
+Project releases and documentation versions are strongly related each other.
+You usually want to have a release ``1.0.3`` with it's own documentation version ``1.0.3`` for that specific release.
+
+When you make a new release of your software,
+you create a new tag in your :abbr:`VCS (Version Control System)` for this release and activate this new version on Read the Docs.
+At this point, Read the Docs will build and publish the new documentation version ``1.0.3`` for us.
```
davidfischer:
```suggestion
At this point, Read the Docs will build and publish the new documentation version ``1.0.3`` for you.
```

You use "you" in the rest of this paragraph so let's be consistent.

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+.. meta::
+   :description lang=en:
+
+      Recommendations about how to do documentation versioning.
+
+Documentation Versioning: Best Practices
+========================================
+
+Project releases and documentation versions are strongly related each other.
+You usually want to have a release ``1.0.3`` with it's own documentation version ``1.0.3`` for that specific release.
```
davidfischer: Django and Flask don't do this. I'd actually say if you're following semver you probably don't need to document below the minor version level (eg. `1.0.x`).

documentation-versioning.rst
davidfischer: I think the key is that if there will be more 1.0.x releases, there should be a release branch and documentation builds off that branch. If you have a typo in the docs or an example in the docs that doesn't work, you just make a new version and the `1.0.x` docs get updated.

It also depends on whether there will be multiple concurrent versions. For example, Django is currently supporting 1.11.x, 2.1.x and 2.2.x and there are large differences between these. Users definitely should be looking at the right docs for their version.

If instead, you just want to have a few old versions of the docs around for users using the old stuff then I think tags are mostly fine. However, again if you're following semver you probably don't need docs for `1.0.3` and `1.0.4`. I think that's worth mentioning in this post!

davidfischer: > if you're following semver you probably don't need docs for 1.0.3 and 1.0.4

I mean you have the docs in VCS but you don't need published docs for both when they're basically exactly the same.

humitos:
Marking this as WIP because I want to revisit and re-write some sections after the feedback that I received and the discussion we have.

humitos:
I revisited this PR twice this month and I'm still stuck thinking what's the recommendation I feel comfortable communicating to the audience. In the end, I found myself wanting **version aliases** combined with **automation rules**:

- tag all the releases following SemVer (`x.y.z`)
- always build latest/stable
- an automated rule will create a version alias `x.y` (dropping `z` patch number) for the latest `x.y.z` tag in that minor `y` version

That will give me these type of versions, for example:

* tags in VCS
  - `0.6.2`
  - `0.6.3`
  - `1.2.0`
  - `1.3.0`
  - `1.3.1`
  - `2.0.0`
* alias on Read the Docs
  - `0.6.x` -> points to `0.6.3`
  - `1.3.x` -> points to `1.3.1`
  - `2.0.x` -> points to `2.0.0`
  - `latest` -> points to `master`
  - `stable` -> points to `2.0.0`

(all the versions that are not alias are `active=False` in my head: 5 are enabled in total for this example)

Instead of putting the effort on the release process, we build a feature around it that makes it automatically for the users.

I really like what @davidfischer said in the last comment. However, I think the size of the project shouldn't be the main difference. I think that even if we are not maintaining `1.3.x` anymore, we'd like users to read the latest *patch version* of that documentation instead of `1.3.0`.

> I'd look at our top 20-30 projects and see what they do. Do they all fit these two strategies?

- pip: latest/stable
- django: release branches (`stable/ + ask us for custom version alias + latest/stable
- flask: call the branches *exactly* `1.1.x`, `1.0.x` and `0.12.x` + `master` version
- godot: release branches `3.2`, `3.1`, `3.0` and `2.1`
- requests: looks unmaintained. However, it looks like a good example of "latest patch version". They have the full SemVer (`2.9.1`) published, but then they released `2.9.2` and it's not published.
- scrapy: release branches exactly named as flask does
- bootstrap-datepicker: publish all full SemVer tags (`1.7.1` and `1.7.0`)

----

Summarizing, I'd like to promote two options whenever that would be possible:

- latest/stable
- automation rules + version alias

The others are too complicated in my opinion, require a specific branch naming in your VCS, point users to old documentation (fixed patch version tag), etc.

documentation-versioning.rst
```diff
@@ -0,0 +1,71 @@
...
+Similarly, if you want to install a new Sphinx extension in all your versions,
+update one of its settings, change the theme,
+or even change a build config in the `.readthedocs.yml`_,
+you will be "blocked" by the tags being static.
+
+
+Recommended versioning strategy
+-------------------------------
+
+**The workflow we recommend for versioning your documentation is having release branches.**
```
astrojuanlu, line 49: I get that this idea effectively encourages projects to follow Trunk Based Development https://trunkbaseddevelopment.com/ but there are other schemes https://www.toptal.com/software/trunk-based-development-git-flow

I wonder how applicable this is for projects not following Trunk Based Development

astrojuanlu:
I'm going to distill some of these ideas in the tutorial https://github.com/sphinx-doc/sphinx/pull/9355/

astrojuanlu:
(Edit, wrong link)

PR #76, opened 2019-08-26
SEO blog post

davidfischer:
* Add a blog about the SEO guide while trying not to just duplicate the guide
* The main audience for this is project maintainers and technical writers who use Sphinx/RTD. The goal is to get them to read the actual SEO guide and hopefully improve their docs.
* The high level goal is this is content marketing.


humitos:
I'm late here but I think it would have been good to make it agnostic of the tool. Reading the title, it seems that this only applies to Sphinx but I didn't find very strict recommendations about how to do it in Sphinx. It's mentioned in the blog post a few times, but that could have been avoided or expanded with MkDocs as well.

I think it could be MkDocs/Sphinx which are the tools we support.

The blog post is really good, though :)

PR #77, opened 2020-03-17
Add CI and fix requirements and broken links

stsewd:
The branch/fork that we were installing ablog from was deleted, so we can't build the blog anymore. The ablog project was moved the sunpy org (https://github.com/abakan-zz/ablog/pull/93#issuecomment-366457676), the latest version to support sphinx 1.7 is `0.9.5`. Updating to sphinx > 1.8 breaks the custom CSS we have for the blog (we should try updating it in another moment anyway).

I also started to use `-W` to treat warnings as errors, and found some broken links and setup CI using github actions.

The blog looks the same as far I was able to see updating to 0.9.5

### DIFF
--- a/ad-blocker-update.rst
+++ b/ad-blocker-update.rst
@@ -8,7 +8,7 @@ Update on Ad Blocking and Acceptable Ads
 ========================================
 
 A few weeks ago, we shared about the
-:doc:`challenge ad blocking presented to our sustainability <ads-and-adblocking>`
+:doc:`challenge ad blocking presented to our sustainability </ads-and-adblocking>`
 and what we were doing about it.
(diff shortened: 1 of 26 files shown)


ad-blocker-update.rst
```diff
@@ -8,7 +8,7 @@ Update on Ad Blocking and Acceptable Ads
 ========================================
 
 A few weeks ago, we shared about the
-:doc:`challenge ad blocking presented to our sustainability <ads-and-adblocking>`
+:doc:`challenge ad blocking presented to our sustainability </ads-and-adblocking>`
```
stsewd: For some reason sphinx gives a warning that it can locate the document when it doesn't start with `/`, so I had to update those.

PR #79, opened 2020-04-23
Blog post about better conda support

humitos:
First draft.

I'm trying to communicate that RTD is better now by telling a little bit of the history behind this. Why it wasn't working before? What we have tried? What we have now?

It mentions all the points that I had in mind at least. It may still require some work on wording tho.

Rendered version at https://readthedocs-blog--79.org.readthedocs.build/better-conda-support/

### DIFF
--- a/better-conda-support.rst
+++ b/better-conda-support.rst
@@ -1,0 +1,45 @@
+.. post:: April 23, 2020
+   :tags: conda, builders, science
+   :author: Manuel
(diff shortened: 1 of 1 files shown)


ericholscher:
@humitos I did a pretty through edit, but tried to keep the content the same, if you want to take a peek 👍

humitos:
I really like the edits :smile:

PR #83, opened 2020-06-16
Embedding Content From Your Documentation, blog post

humitos:
We have covered most of this feature in our guide at
https://docs.readthedocs.io/en/stable/guides/embedding-content.html, so I don't
think there is too much to add here. However, I think the blog is the right
place to promote this feature and share a little of its story.

I'd like to have an example of in-app UI tooltip, but the tool we built does not
look nice and I don't think we want to promote that as is.

> rendered version: https://readthedocs-blog--83.org.readthedocs.build/embedding-content-from-your-docs/


agjohnson:
The embed API is not stable on commercial, so we should probably get to some level of parity on commercial before we start talking about embedding as a feature. It's currently only a feature of community and so commercial users will be confused by this news.

humitos:
> The embed API is not stable on commercial

What do you mean by this? We deployed the Embed API some time ago in .com and I remember that it was working when we did some QA. I will double check this, anyways.

humitos:
I just built sphinx-hoverxref's documentation and it works: https://docs.ops.verbthenouns.com/projects/read-the-docs-sphinx-hoverxref/en/corporate-api-endpoint/

(it builds a different branch than latest to use the proxied API instead of community URL as it was doing)

ericholscher:
That branch isn't active for me.

ericholscher:
I had a bunch of feedback here, so I went ahead and just took a swing at re-writing it. If we're going to be playful with the tone, we need to be really explicit and clear about what we're saying, because it can be confusing to people otherwise. I tried to keep a little of the playfulness, but keep it more obvious what we were talking about.

ericholscher:
Before we ship this, we should confirm the API is something we want to support. Currently we're returning `headers`, but I thought we didn't want to?

astrojuanlu:
Should we hold this off until https://github.com/readthedocs/readthedocs.org/pull/8039 is merged?

humitos:
I'm fine publishing this. The current API has to be maintained anyways, so I'd say it's fine if people start using it. However, the most important goal of this post to me is the promotion of the Sphinx extension.

Waiting for the new Embed API + JS client will be too much time, IMHO.

embedding-content-from-your-docs.rst
```diff
@@ -0,0 +1,71 @@
...
+
+.. _sphinx-hoverxref: https://sphinx-hoverxref.readthedocs.io/
+.. _reading the docs: https://sphinx-hoverxref.readthedocs.io/
+
+Inline help on our application website
+--------------------------------------
+
+Adding too much copy in your application can be overwhelming to users,
+meaning they don't read it.
+We try to keep the application copy minimal, and link out to documentation when appropriate.
```
astrojuanlu, line 54: **question:** What does "the application copy" mean in this context?

embedding-content-from-your-docs.rst
```diff
@@ -0,0 +1,71 @@
...
+This pattern has worked fine over time, but it has the same problem of context switching to a completely different page.
+
+Since we've started working on a new re-design of our entire application UI (more on this coming soon!),
+we added the usage of the Embed API to add this extra documentation.
```
astrojuanlu, line 58: **suggestion**: Rephrase to avoid saying "add" twice in the same sentence?

embedding-content-from-your-docs.rst
```diff
@@ -0,0 +1,71 @@
...
+In summary
+----------
+
+If you liked the Embed API and you would like to embed content from your documentation anywhere,
+you can `read its full documentation`_ to learn more about it.
+
+.. _read its full documentation: https://docs.readthedocs.io/en/stable/guides/embedding-content.html
+
+We think it's a super powerful feature and we encourage you to use the extension we built on top of it,
+or even better, build your own and share them with us!
```
astrojuanlu, line 71: **nitpick:** The tone looks a bit informal here.

I wonder if we should make it a bit more formal, in line with other blog posts.

humitos:
I think we can close this PR since the work has been done in other places already.

astrojuanlu:
Wow, somehow I totally forgot about this blog post. We should have published it earlier. Thanks @humitos

PR #90, opened 2021-01-11
Updates on Project Privacy Levels

stsewd:
We need to

- Decide on a date to make the change
- Optionally make it obvious if a project is private in the UI in .org
- Optionally merge https://github.com/readthedocs/readthedocs.org/pull/7608
- Mail all users with private/protected projects in .org about this change (maybe twice?)
- Do a data migration
  - For .org private/protected to public
  - For .com protected to private

Try it to keep it short and clear, hopefully isn't confusing :D

### DIFF
--- a/project-privacy-levels-update.rst
+++ b/project-privacy-levels-update.rst
@@ -1,0 +1,44 @@
+.. post:: Jan 14, 2021
+   :tags: privacy-levels, deprecation
+   :author: Santos
+   :location: CUE
+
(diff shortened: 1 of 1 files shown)


ericholscher:
@stsewd I think we did this work, but we don't need the blog post right? Should we close it?

stsewd:
> @stsewd I think we did this work, but we don't need the blog post right? Should we close it?

yeah, I have forgotten I wrote this p:

PR #92, opened 2021-01-20
Announcing scheduled maint. for AWS migration on community

ericholscher:
What it says on the tin.
Mostly just to notify users that the dashboard will be read-only during this time.

### DIFF
--- a/aws-migration.rst
+++ b/aws-migration.rst
@@ -7,7 +7,7 @@
 Read the Docs Community downtime due to migrations to AWS
 =========================================================
 
-We wanted to make you aware that on Friday, January 29 at 1pm PST (4pm EST, 20:00 UTC),
+We wanted to make you aware that on Friday, February 12th at 1pm PST (4pm EST, 20:00 UTC),
 Read the Docs Community (readthedocs.org) will be having a **scheduled dashboard downtime of approximately 4 hours**.
 
 To ensure minimal impact for our users,
--- a/aws-migration.rst
+++ b/aws-migration.rst
@@ -1,4 +1,4 @@
-.. post:: Jan 25, 2021
+.. post:: Feb 5, 2021
    :tags: aws, downtime, infrastructure
    :author: Eric
    :location: BND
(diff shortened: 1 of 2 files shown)


humitos (approved):

aws-migration.rst
```diff
@@ -0,0 +1,30 @@
+.. post:: Jan 21, 2021
+   :tags: aws, downtime, infrastructure
+   :author: Eric
+   :location: BND
+
+
+Read the Docs Community downtime due to migrations to AWS
+=========================================================
+
+We wanted to make you aware that on Friday, January 29 at 12:01pm PDT (4:00pm EDT, 20:00 UTC),
```
davidfischer, line 10:
```suggestion
We wanted to make you aware that on Friday, January 29 at 12:01pm PST (4:00pm EST, 20:00 UTC),
```

aws-migration.rst
```diff
@@ -0,0 +1,30 @@
...
+
+We wanted to make you aware that on Friday, February 12th at 1pm PST (4pm EST, 20:00 UTC),
+Read the Docs Community (readthedocs.org) will be having a **scheduled dashboard downtime of approximately 4 hours**.
+
+To ensure minimal impact for our users,
+we are performing this upgrade during a Friday afternoon
+which is one of our lowest usage periods.
+
+During this maintenance window, **documentation will continue to be online**
+but new documentation builds will not trigger and the Read the Docs dashboard will be read-only.
```
agjohnson, line 18:
```suggestion
but new documentation builds will not trigger and the Read the Docs dashboard will be unavailable.
```

ericholscher: I thought we'd tested this with a read-only user in the DB? Seems pretty simple to apply while we're updating things, but yea it might break the dashboard. 🤷

agjohnson: Ah yes, so internally we'll use a read-only connection to the database to keep docs up, but the dashboard will have to be completely down as it will throw exceptions everywhere with a read-only db connection

PR #96, opened 2021-03-03
Add the open position to the blog sidebar

davidfischer:
I purposefully used absolute links because relative links might break on some pages (eg. https://blog.readthedocs.com/archive/2017/)

### DIFF
--- a/_themes/rtd-blog/social.html
+++ b/_themes/rtd-blog/social.html
@@ -14,8 +14,16 @@ Sign up for our blog and we'll send you information about Sphinx and Read the Do
     </div>
 </form>
 </div>
-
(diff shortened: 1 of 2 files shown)


humitos (approved):

ericholscher:
This is good, but I don't think our blog has much value by itself. I think we want a landing page that lists all open positions that we can link to from the main RTD site. This would have the jobs as the primary content, not something lost in the sidebar.

ericholscher:
I pushed up an additional jobs landing page for us to link to.

davidfischer:
> I pushed up an additional jobs landing page for us to link to.

:+1:

PR #100, opened 2021-04-06
Frontend job update

ericholscher:
This mainly tries to make the job more welcoming to folks without as much experience or confidence.
Definitely open to feedback here around ways to make it even more welcoming.

I didn't totally refactor it,
but tried to soften the language and show more support.


astrojuanlu (approved):
I think this is definitely less intimidating than before, and clearer as well 👍🏽

PR #109, opened 2021-06-09
Add blog post about sphinx-hoverxref and intersphinx support

astrojuanlu:
As per title 😄 

As I did not see other in-depth mentions of `sphinx-hoverxref` in our blog, I used this opportunity to introduce the extension a bit more formally.

🤖 Edit: Rendered version at https://readthedocs-blog--109.org.readthedocs.build/hoverxref-intersphinx/ (joking, I'm a human, but I would love a bot that does this!)


astrojuanlu:
I see that there are two tags that didn't get the corresponding links, not sure why 🤔

hoverxref-intersphinx.rst
```diff
@@ -0,0 +1,108 @@
...
+can be distracting, since opening them might need either a context switch
+or keeping a new browser tab for a later moment.
+
+Here is where `sphinx-hoverxref`_ comes to the rescue:
+it scans the cross-references and attaches floating windows to them
+(either *tooltips* or *modal* dialogues)
+that embed the contents of the page the xrefs point to.
+This way, when the user is interested in a particular topic,
+they can hover the mouse over the link
+and display the floating window instead.
```
humitos, line 52: nit: I understand that this "instead" references to the previous paragraph, right? The first time I read it I didn't do the link and I felt that a part was missing. Like, "... instead of opening the links on tabs"

So, does it make sense to make these two paragraphs just one? or continue the sentence after the "instead"?

hoverxref-intersphinx.rst
```diff
@@ -0,0 +1,108 @@
...
+This extension is an essential part
+of the work we are doing to improve interoperability of documentation in general,
+and scientific documentation in particular,
+thanks to the `CZI grant </czi-grant-announcement>`_ we received last year.
+
+.. figure:: img/sphinx-hoverxref-intersphinx.gif
+   :width: 100%
+   :align: center
+   :alt: sphinx-hoverxref displaying a modal window embedding an external project
+   :target: _images/sphinx-hoverxref-intersphinx.gif
```
humitos, line 27: Does it make sense to change this GIF with a real-life example that we can link to the project as well? This was just an example that I record to show to the team, but it doesn't exist.

Also, this particular case (intersphinx feature) may need an extra explanation since it's not easy to realize that the content displayed in the modal comes from a completely different project.

humitos: Oh, I see that you are introducing intersphinx later on in the text. So, we could add the GIF in that section with an explanation.

Here, in the beginning, we could just add the most basic case, which is the small "tooltip" style to reference an internal link.

hoverxref-intersphinx.rst
```diff
@@ -0,0 +1,108 @@
...
+   :width: 100%
+   :align: center
+   :alt: sphinx-hoverxref displaying a tooltip including an equation
+   :target: _images/sphinx-hoverxref-math.png
+
+Support for ``intersphinx``
+---------------------------
+
+Version 0.6b1 of `sphinx-hoverxref`_ introduces opt-in support for Intersphinx linking.
+``sphinx.ext.intersphinx`` is a Sphinx extension
```
humitos, line 68: Would be good to have a link to the docs here.

hoverxref-intersphinx.rst
```diff
@@ -0,0 +1,108 @@
...
+that generates automatic links to the documentation of objects in other projects,
+and now floating windows are shown for them too.
```
humitos, line 70:
```suggestion
and now it's possible to show floating windows on them too.
```

I didn't like the original one 😄 --but I don't have a great suggestion either 😞 . I just wanted to specify that it's not by default.

hoverxref-intersphinx.rst
```diff
@@ -0,0 +1,108 @@
...
+   hoverxref_intersphinx = [
+      "astropy",
+      "numpy",
+      "scipy",
+      "matplotlib",
+   ]
+
+Moreover, the look & feel of the current project is preserved
+even if the target documentation has a different theme or colors,
+which makes the result more visually consistent.
```
humitos, line 87: While this is true and can be seen as a feature, it's not something we can support/guarantee. I think it's fine to leave it as is, tho.

There are ways to add some simple CSS in case the look & feel is different, but I don't think we should go too technical here, either.

humitos (commented):
Looks great! I left some comments/suggestions to consider.

hoverxref-intersphinx.rst
astrojuanlu: Fair enough! I can switch the order of the images. I originally put the animated GIF first because perhaps it captivates the reader attention :)

hoverxref-intersphinx.rst
astrojuanlu: I prefer your suggestion!

hoverxref-intersphinx.rst
astrojuanlu: Yeah, it's more of a realization of the current state of things, rather than a committment.

hoverxref-intersphinx.rst
humitos: The first one could still be a GIF with the small tooltip example 😉

hoverxref-intersphinx.rst
astrojuanlu: You're right, good catch! I continued the sentence

hoverxref-intersphinx.rst
astrojuanlu: Sure thing, although I am *really* annoyed by the lack of nested markup in reStructuredText (I would like this to be a link _and_ typewriter font). So much for the "superior" language 😒 (there, I said it)

astrojuanlu: Changed 👍🏽

hoverxref-intersphinx.rst
astrojuanlu: Swapped the images. If you still think we should record a new animation, let me know.

astrojuanlu:
Thanks @humitos for the review! Addressed the comments 👍🏽

hoverxref-intersphinx.rst
```diff
@@ -0,0 +1,114 @@
...
+.. meta::
+   :description lang=en:
+      New release of sphinx-hoverxref, our Sphinx extension to add tooltips to cross references,
+      with support for intersphinx.
+
+New release of sphinx-hoverxref with support for intersphinx
+============================================================
+
+We have released version 0.6b1 of `sphinx-hoverxref`_,
+a Sphinx extension that shows a floating window
```
ericholscher, line 15: I feel like we should define a standard name for this. Wikipedia calls it `Page previews` which seems like a reasonable name to copy. 

I think `preview` is probably useful to call this instead of floating window, since it has more sematic meaning. You're actually getting to "preview" or "reference" the content without leaving the page, whereas floating window is a technical description.

hoverxref-intersphinx.rst
```diff
@@ -0,0 +1,114 @@
...
+
+   sphinx-hoverxref displaying a modal window embedding an external project
+
+To enable it, the user needs to specify
+which projects from the `intersphinx_mapping <https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_mapping>`_
+will be scanned by ``sphinx-hoverxref``, using a new configuration value:
+
+.. code-block:: python
+
+   hoverxref_intersphinx = [
```
ericholscher, line 84: I still feel like this name is not great, and should be more explicit. I feel like this code example actually needs a comment to explain _what_ the content of this list is:

```suggestion
   # Values of the intersphinx_mapping keys that you want to enable hoverxref on
   hoverxref_intersphinx = [
```

hoverxref-intersphinx.rst
```diff
@@ -0,0 +1,114 @@
...
+- `The Grill`_, an extension for digital content creation with Maya, Houdini, and other 3D modeling applications,
+  and references another project called `naming <https://naming.readthedocs.io/>`_.
+- `poliastro`_, a Python library for interactive Astrodynamics
+  that makes cross-references to `Astropy <https://docs.astropy.org/>`_ and other projects.
+
+We are excited to see what our users can do with `sphinx-hoverxref`_,
+and will continue working on it in the future.
+
+Considering using Read the Docs for your next Sphinx or MkDocs project?
+Check out `our documentation <https://docs.readthedocs.io/>`_ to get started!
```
ericholscher, line 111: It's a small thing, but I love that we have this in every blog post :)

It should probably be part of the blog theme, and on the bottom of every page automatically in the future.

hoverxref-intersphinx.rst
```diff
@@ -0,0 +1,114 @@
...
+
+   At the moment, the target project needs to be hosted on Read the Docs,
+   which is a limitation that we hope to remove in the future.
+
+You can draw inspiration from some community projects that are using this feature already:
+
+- `The Grill`_, an extension for digital content creation with Maya, Houdini, and other 3D modeling applications,
+  and references another project called `naming <https://naming.readthedocs.io/>`_.
+- `poliastro`_, a Python library for interactive Astrodynamics
+  that makes cross-references to `Astropy <https://docs.astropy.org/>`_ and other projects.
```
ericholscher, line 105: Social proof is 💯

hoverxref-intersphinx.rst
```diff
@@ -0,0 +1,114 @@
...
+   :width: 80%
+   :align: center
+   :alt: sphinx-hoverxref displaying a tooltip including an equation
+   :target: _images/sphinx-hoverxref-math.png
+
+   sphinx-hoverxref displaying a tooltip including an equation
+
+.. _sphinx-hoverxref: https://sphinx-hoverxref.readthedocs.io/
+
+Embedding documentation fragments
```
ericholscher, line 33: This is another place where `previews` is probably more descriptive. Talk about the _value_ to the user, not the technical implementation. Nobody cares about fragments, or even knows what it means in this context. 

```suggestion
Embedding reference previews
```

Not 100% sold on this, but feels better.

ericholscher (approved):
Glad this got written up -- we're actually slowly talking about the work we're doing 🎉

hoverxref-intersphinx.rst
astrojuanlu: Good catch 👍🏽

hoverxref-intersphinx.rst
astrojuanlu: ❤️

hoverxref-intersphinx.rst
astrojuanlu: 😊  Yes, it feels like a great call to action!

> It should probably be part of the blog theme, and on the bottom of every page automatically in the future.

Yeah, I was thinking about it yesterday! Somehow I wouldn't want to fight against it if I don't want to show it for a particular blog post. But I think it would be generally a good idea.

hoverxref-intersphinx.rst
astrojuanlu: What about "content preview"?

astrojuanlu: (I went ahead and renamed it to "content preview", let me know if you still prefer "reference preview" 👍🏽)

hoverxref-intersphinx.rst
astrojuanlu: Good point. This is something to consider for the `sphinx-hoverxref` documentation as well.

astrojuanlu:
Addressed some style comments and reworked the images so they are more meaningful. @ericholscher already approved, so if there are no other comments I'll merge & announce on social media in a few minutes.

hoverxref-intersphinx.rst
ericholscher: That works for me.

hoverxref-intersphinx.rst
ericholscher: Yea, definitely should backport these changes there.

humitos: Yeah, I've been there already: https://sphinx-hoverxref.readthedocs.io/en/latest/usage.html#id2 🙈

PR #128, opened 2021-09-16
Change newsletter form

astrojuanlu:
Change newsletter form to use MailerLite.

Screenshot 2021-09-16 at 11-48-20 Recent posts

- I added a mandatory checkbox for confirmation
- I added a reCAPTCHA, it can be removed
- I made some style tweaks
- Subscribers in this form go to our `Mailchimp Import` list, it can be changed

### DIFF
--- a/_themes/rtd-blog/social.html
+++ b/_themes/rtd-blog/social.html
@@ -4,9 +4,9 @@
     <div class="ml-form-align-center">
       <div class="ml-form-embedWrapper embedForm">
         <div class="ml-form-embedBody ml-form-embedBodyDefault row-form">
-          <div class="ml-form-embedContent" style="">
+          <div class="ml-form-embedContent">
             <h3 style="margin-left: 0">Subscribe to our mailing list</h3>
-            <p>Sign up for our blog and we'll send you information about Sphinx and Read the Docs on a regular basis.</p>
+            <p>Sign up for our blog and we'll send you news and updates about Sphinx and Read the Docs on a regular basis.</p>
           </div>
           <form class="ml-block-form" action="https://static.mailerlite.com/webforms/submit/p8b7z2" data-code="p8b7z2" method="post" target="_blank">
             <div class="ml-form-formContent">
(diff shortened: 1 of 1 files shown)


stsewd:
We also need to update this https://github.com/readthedocs/readthedocs.org/blob/0e5818aaa4b964a2eb3f1dc0bcddcdc7d318a3b9/readthedocs/templates/base.html#L137-L164

_themes/rtd-blog/social.html
```diff
@@ -1,19 +1,53 @@
...
+          <form class="ml-block-form" action="https://static.mailerlite.com/webforms/submit/p8b7z2" data-code="p8b7z2" method="post" target="_blank">
+            <div class="ml-form-formContent">
+              <div class="ml-form-fieldRow ml-last-item">
+                <div class="ml-field-group ml-field-email ml-validate-email ml-validate-required">
+                  <input aria-label="email" aria-required="true" type="email" class="form-control" data-inputmask="" name="fields[email]" placeholder="email address" autocomplete="email" style="width: 90%">
+                </div>
+              </div>
+            </div>
+            <div class="ml-form-checkboxRow ml-validate-required">
+              <label class="checkbox"> <input type="checkbox"> <div class="label-description"> <p></p> <p>I wish to subscribe to the Read the Docs newsletter to receive news and updates. I&nbsp;know I&nbsp;can unsubscribe at any time.</p> <p></p> </div> </label>
```
ericholscher, line 20: What is the goal here? I don't think this really needs to be a checklist, but probably copy under the subscribe link?

_themes/rtd-blog/social.html
```diff
@@ -1,19 +1,53 @@
...
+            </div>
+            <div class="ml-form-recaptcha ml-validate-required" style="float:left">
+              <style type="text/css">
+                .ml-form-recaptcha{margin-bottom:20px}.ml-form-recaptcha.ml-error iframe{border:solid 1px red}@media screen and (max-width:480px){.ml-form-recaptcha{width:220px!important}.g-recaptcha{transform:scale(.78);-webkit-transform:scale(.78);transform-origin:0 0;-webkit-transform-origin:0 0}}
+              </style>
+              <script src="https://www.google.com/recaptcha/api.js"></script>
+              <div class="g-recaptcha" data-sitekey="6Lf1KHQUAAAAAFNKEX1hdSWCS3mRMv4FlFaNslaD"></div>
+            </div>
+            <input type="hidden" name="ml-submit" value="1">
+            <div class="ml-form-embedSubmit">
```
ericholscher, line 30: Can we move the Captcha to after submitting the form? Feels really weird & distracting to have in the sidebar for all users.

ericholscher (commented):
This looks good, but I think it's a bit overkill. I assume people are submitting the form know they are signing up for the mailing list?

_themes/rtd-blog/social.html
astrojuanlu: Unfortunately, MailerLite doesn't support that, I've been told. We didn't have a reCAPTCHA previously, so I suppose we can continue without it. Mildly annoying though.

ericholscher: Seems we could just have a form that goes to the next page with more details (eg. our own little landing page). I say let's keep it off for now, and we can iterate. I'd like to get this shipped today.

_themes/rtd-blog/social.html
astrojuanlu: Simplified this a bit

astrojuanlu:
Simplified the form a bit:

Screenshot 2021-09-20 at 09-49-07 Recent posts

_themes/rtd-blog/social.html
humitos: IIRC, this is required by GDPR. Just copy is not enough. After a quick google search, many sites mention something like

> Consent requires a positive opt-in. Don't use pre-ticked boxes or any other method of default consent.

Also, I remembered that we had to add this checkbox to our PyCamp web page.

astrojuanlu:
Changed also on the community site https://github.com/readthedocs/readthedocs.org/pull/8509

_themes/rtd-blog/social.html
ericholscher: If it's required for GDPR, we should do an opt-in email after they have subscribed. A small button on the page is not going to meaningfully be an opt-in (since it can be done by anyone on the web page).

ericholscher:
Looks good. I do think a capcha (or double opt-in via email) is going to be a good practice going forward, but let's just get it to "not broken" on this PR :)

ericholscher (approved):

PR #147, opened 2021-12-15
Add blog post about Sphinx and Markdown

astrojuanlu:
I have experimented with different approaches, and finally settled with this. Talking only about our conferences seemed a weird thing to do, so instead I framed them as part of our larger efforts to promote Markdown, which are also in line with the CZI grant. Let me know what you think @ericholscher

### DIFF
--- a/sphinx-markdown-2021.rst
+++ b/sphinx-markdown-2021.rst
@@ -1,11 +1,13 @@
-.. post:: Dec 15, 2021
-   :tags: media, talk, conference
+.. post:: Dec 16, 2021
+   :tags: media, talk, conference, markdown, commonmark
    :author: Juan Luis
    :location: MAD
 
 .. meta::
    :description lang=en:
-      ...
+      During 2021 we have promoted the use of Markdown in Sphinx in various ways.
+      In this post we show how you can start using it in your project today,
+      and give some extra pointers.
 
 Sphinx and Markdown around the world in 2021
 ============================================
--- a/sphinx-markdown-2021.rst
+++ b/sphinx-markdown-2021.rst
@@ -1,4 +1,4 @@
-.. post:: Dec 16, 2021
+.. post:: Dec 20, 2021
    :tags: media, talk, conference, markdown, commonmark
    :author: Juan Luis
    :location: MAD
--- a/sphinx-markdown-2021.rst
+++ b/sphinx-markdown-2021.rst
@@ -15,11 +15,11 @@ Sphinx and Markdown around the world in 2021
 Read the Docs has been committed to improving the accessibility
 and user experience of Sphinx since the start,
 and that includes the markup language in which the documentation is written.
-After carefully listening to users
-that wanted to use Markdown in Sphinx as well as reStructuredText,
-:doc:`we announced recommonmark more than six years ago </adding-markdown-support>`
+Years ago, after carefully listening to users,
+:doc:`we announced recommonmark </adding-markdown-support>`
 to help bridging the immense popularity of Markdown
 with the powerful capabilities of Sphinx.
+(It is now deprecated in favor of MyST - keep reading to know more.)
 
 It is no surprise that Markdown is in such demand:
 thanks in large part to the huge popularity of GitHub,
--- a/sphinx-markdown-2021.rst
+++ b/sphinx-markdown-2021.rst
@@ -31,15 +31,36 @@ especially `the Executable Book Project <https://executablebooks.org>`_,
 we have devoted lots of resources to help consolidate Markdown
 as a compelling markup language for Sphinx.
 
+Writing Markdown in Sphinx
+--------------------------
+
 To that end, we recognized the potential of MyST as a successor of recommonmark,
 and :doc:`deprecated the latter in favor of MyST-Parser </newsletter-april-2021>`.
-Luckily the migration was trivial for most projects,
-but we still wanted to spread the word about
-the best practices for creating Sphinx projects on Read the Docs using MyST Markdown.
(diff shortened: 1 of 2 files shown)


astrojuanlu:
🤖 Rendered version: https://readthedocs-blog--147.org.readthedocs.build/sphinx-markdown-2021/

sphinx-markdown-2021.rst
```diff
@@ -0,0 +1,115 @@
...
+============================================
+
+Read the Docs has been committed to improving the accessibility
+and user experience of Sphinx since its beginnings,
+and that includes the markup language in which the documentation is written.
+After carefully listening to users
+that wanted to use Markdown in Sphinx as well as reStructuredText,
+:doc:`we announced recommonmark more than six years ago </adding-markdown-support>`
+to help bridging the immense popularity of Markdown
+with the powerful capabilities of Sphinx.
```
humitos, line 22: I found this sentence confusing when I read it. At first, I thought we listen to our users in the last months, but then I realized that it's taking about 6 years ago. However, I've already clicked on the blog post and I was reading it 😄 

If possible, I'd change the order of it to mention the time at the beginning of the sentence to immediately contextualize the reader. Something like,

```
Years ago, after carefully listening to users ... we announced recommonmark
```

Besides, I think it would be good to mention in the same paragraph that it's currently deprecated, something like:

```
which is now, deprecated in favor of MyST (keep reading to find out more about how to use it / migrate)
```

sphinx-markdown-2021.rst
```diff
@@ -0,0 +1,115 @@
...
+we have devoted lots of resources to help consolidate Markdown
+as a compelling markup language for Sphinx.
+
+Writing Markdown in Sphinx
+--------------------------
+
+To that end, we recognized the potential of MyST as a successor of recommonmark,
+and :doc:`deprecated the latter in favor of MyST-Parser </newsletter-april-2021>`.
+To :doc:`start writing Markdown in Sphinx <sphinx:usage/markdown>`,
+you need to follow three steps:
```
humitos, line 42: Maybe we can remove the deprecation notice from here and put it only on the initial paragraph instead.

Here we could just talk about MyST without mentioning recommonmark. If we want to highlight that recommonmark is deprecated, we could have a specific section for this, tho. I think that mentioning this deprecation here is out of the scope of the section's title and adds confusion only.

Actually, do we want to mention recommonmark in this post at all? 😄 . Maybe the focus on MyST + Sphinx + conferences is 💯 without recommonmark.

sphinx-markdown-2021.rst
```diff
@@ -0,0 +1,115 @@
...
+* Start writing content in ``*.md`` files!
+
+You can find more information in the :doc:`MyST-Parser documentation <myst-parser:index>`.
+
+Talks and workshops at conferences
+----------------------------------
+
+Luckily the migration to MyST-Parser was trivial for most projects,
+but we still wanted to spread the word about
+the best practices for creating Sphinx projects on Read the Docs using MyST Markdown.
```
humitos, line 65: Now, after reading this paragraph, I seems that migrating users from recommonmark to MyST was part of this job and now, having a specific section commenting that recommonmark was deprecated in favor or MyST makes more sense.

humitos (commented):
The post looks great to me!

I added some small suggestions/comments that are not mandatory but could be considered. Keep in mind that _I don't know how to do the work to make these changes_, but I think the structure/mention around recommonmark could be improved.

So, if you think what I'm saying makes sense and you find a way to do it 👍🏼

sphinx-markdown-2021.rst
astrojuanlu: Since you made the earlier comments before reading the whole post, I prefer to wait on @ericholscher to do the proofreading. Thanks anyway @humitos !

sphinx-markdown-2021.rst
```diff
@@ -0,0 +1,115 @@
...
+   :description lang=en:
+      During 2021 we have promoted the use of Markdown in Sphinx in various ways.
+      In this post we show how you can start using it in your project today,
+      and give some extra pointers.
+
+Sphinx and Markdown around the world in 2021
+============================================
+
+Read the Docs has been committed to improving the accessibility
+and user experience of Sphinx since its beginnings,
```
ericholscher, line 16:
```suggestion
and user experience of Sphinx since the start,
```

reads a bit better.

sphinx-markdown-2021.rst
ericholscher: Yea, 👍 to mentioning that it's deprecated in the same sentence.

sphinx-markdown-2021.rst
```diff
@@ -0,0 +1,115 @@
...
+and that includes the markup language in which the documentation is written.
+After carefully listening to users
+that wanted to use Markdown in Sphinx as well as reStructuredText,
+:doc:`we announced recommonmark more than six years ago </adding-markdown-support>`
+to help bridging the immense popularity of Markdown
+with the powerful capabilities of Sphinx.
+
+It is no surprise that Markdown is so demanded:
```
ericholscher, line 24:
```suggestion
It is no surprise that Markdown is in such demand:
```

sphinx-markdown-2021.rst
ericholscher: I think we mention it at the top, but talk mostly about the future.

ericholscher (approved):
This looks good to me. I agree we should mention that recommonmark is deprecated in the beginning. I don't feel strongly about the narrative around the migration, I think it's useful either way.

sphinx-markdown-2021.rst
astrojuanlu: Done ✔️

PR #162, opened 2022-05-04
User-defined build jobs

humitos:
Post announcing the new config key `build.jobs`.

Rendered version 🤖 : https://readthedocs-blog--162.org.readthedocs.build/user-defined-build-jobs/

Related to #160

PR #163, opened 2022-05-04
Update requirements

humitos:
Sphinx 2.4.4 does not have `:caption:` when using `.. code::`. I wanted this for #162 

> This is a test to check if updating requirements breaks something.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-Sphinx==2.4.4
+Sphinx==4.5.0
 sphinxext-opengraph==0.4.2
 docutils<0.18
 jinja2<3.1.0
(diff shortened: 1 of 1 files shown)


humitos:
@ericholscher it seems it works fine. Do you know why we haven't updated ablog and Sphinx? I don't see any comments in the requirements blocking this upgrade.

ericholscher:
Not sure -- as long as it works it seems fine? 🤷

agjohnson:
Note: these blog posts will eventually be ported to Pelican. It might be worth keeping in mind the capabilities there when writing new blog posts. All Sphinx authoring features need to be dropped eventually, including intersphinx, etc.

humitos:
OK. I removed the `:caption:` from my blog post. So, I don't require upgrading Sphinx anymore. Since we are moving to a different doctool, I don't think it makes sense to add unsupported things in the new blog posts 👍🏼

PR #177, opened 2022-09-12
Blog post about Telemetry database and Metabase

### LINKED ISSUE #175: Blog post about metabase graphs of RTD/Sphinx usage
We have done lots of interesting data collection in our metabase research. It would be neat to publish this, and share some of this data with the community. Not a ton of direction here, but curious what we can find in the data!

A couple ideas:

* What are the most popular sphinx extensions?
* Graphs of version usage over time. How do Sphinx/mkdocs/etc. versions get updated after a release?

This could also be a really great Twitter thread to start with, then migrate it into a blog post as we get engagement there.

humitos:
Initial content for the blog post.

- I didn't dig up too much into analyzing each of the graph that I'm showing here. I think it's fine for now to show a generalized view of the usage we will do with this data.
- We can write a specific blog post for a particular graph if we want to in the near future. Analyzing and explaining what we have found.
- I left out a section called "Problems we've found so far" because I think it was too much for it. That section that talks about:
  - db replica and slow queries
  - materialized views to solve the problem of slow queries
  - not being able to order results by SemVer versions
  - inconsistent results and the creation for the "active project" concept
  - mention ``dblink`` as a potential solution for cross database queries

📚 Blog post preview 📚 : https://readthedocs-blog--177.org.readthedocs.build/knowing-more-about-ourselves/

Closes #175


ericholscher (approved):
This is great! I went ahead and did a copy edit of the post and committed it back for ease of updating. I think overall this is a great introduction to what we're doing.

I'd love to see a bit more clarity on exactly what data we're storing. Is it just Python packages & config file data?
 
> As you can see in the examples shown above,
> there are lot of insights that we can get from the data we've been collecting.

We should probably share a couple of these at least. We don't need to go overboard, but this seems like a tease without any interpretation.

> We can write a specific blog post for a particular graph if we want to in the near future. Analyzing and explaining what we have found.

Agreed. This is a great introduction, and shows off some of our cool graphs. Exactly what it should include! 💯 

> I left out a section called "Problems we've found so far" because I think it was too much for it. 

Makes sense. We can go more into depth in a future post, but I like keeping them shorter & focused, and it gives us an obvious post to write in the future. 

If you already have that content drafted, you should throw a PR up with it so it doesn't get lost.

knowing-more-about-ourselves.rst
```diff
@@ -8,58 +8,63 @@
...
-Knowing more about ourselves
-============================
+Knowing more about how people use our service
+=============================================
 
-Read the Docs generates a lot of data on it owns.
+Read the Docs generates a lot of data.
 We are the largest documentation platform out there,
-with hundred thousands projects using our product every day to host their documentation.
+with hundreds of thousands of projects using our product every day to host their documentation.
```
ericholscher, line 16: This still reads weirdly, but I couldn't think of an obvious way to state it better. Maybe:

`with 250,000+ projects`

knowing-more-about-ourselves.rst
```diff
@@ -0,0 +1,155 @@
...
+
+VCS types used by projects with successful builds in last year
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. image:: /img/project-vcs-types.png
+   :align: center
+   :width: 100%
+
+
+Docker image used in the last 15 days
```
ericholscher, line 104: What is `default` here? Also, is `ubuntu-latest` a real option, or an error?

knowing-more-about-ourselves.rst
humitos: I don't have a preference here. Both of them are readable to my level of English 😄

knowing-more-about-ourselves.rst
```diff
@@ -0,0 +1,155 @@
...
+
+With our new data collections project,
+we were able to collect a lot of new, interesting data.
+This new data allows us to answer a whole new set of questions,
+and is mainly related to the build process
+The new information that we're collecting includes:
+
+* Python dependencies installed during a documentation build
+* Apt packages installed prior to building docs
+* The entire contents of the `Read the Docs configuration file <https://docs.readthedocs.io/en/stable/config-file/index.html>`_
```
humitos, line 51: Now we are using a bullet list, we can add more things here if we want:

```suggestion
* OS and Python version used to build the documentation
* Python and Conda dependencies installed during a documentation build
* Apt packages installed prior to building docs
* The entire contents of the `Read the Docs configuration file <https://docs.readthedocs.io/en/stable/config-file/index.html>`_
```

I didn't include them original because it was just a paragraph and I they weren't too relevant. I'm fine if we skip them, tho... They aren't too relevant right at the moment.

knowing-more-about-ourselves.rst
humitos: `default` means the user did not specify anything in their config file. IIRC, on config file v2, this is `latest`.

`ubuntu-latest` is an invalid image and the build should have failed. I'm not sure why it's listed here. I didn't notice it when taking the screenshot. That's wrong and it shouldn't be there. I will take a look at the query and fix it. Thanks!

humitos: Hrm... I checked the query. The user used `build.os: ubuntu-latest`, and the config was translated finally to `build.image: readthedocs/build:latest`. That's why it succeed. However, this is an error and we should raise a build error or similar.

I will remove it from this query for now because I don't want to communicate `ubuntu-latest` on the blog post since it's an invalid option.

humitos:
@ericholscher 

> I went ahead and did a copy edit of the post and committed it back for ease of updating.

Thanks! I like the changes 👍🏼 

> I'd love to see a bit more clarity on exactly what data we're storing. Is it just Python packages & config file data?

Mainly, yes. I updated the bullet list with some more data we are storing. The code with the data structure is at https://github.com/readthedocs/readthedocs.org/blob/46365964a12c180021942c9ae8e803ae25c2f265/readthedocs/telemetry/models.py#L22

> We should probably share a couple of these at least. We don't need to go overboard, but this seems like a tease without any interpretation.

OK. I will try to add a small paragraph on each of the plots to describe how we are reading them or what they are being useful for so far.

> If you already have that content drafted, you should throw a PR up with it so it doesn't get lost.

No. I started writing those bullets that I've put on the description and I realized it was good content, but it would make the general reading more complex without adding too much value for most of non-technical readers.

humitos:
> OK. I will try to add a small paragraph on each of the plots to describe how we are reading them or what they are being useful for so far.

@ericholscher I've done this! Let me know what you think. To me, it's ready to merge and publish it.

knowing-more-about-ourselves.rst
```diff
@@ -0,0 +1,206 @@
+.. post:: September 13, 2022
+   :tags: metabase, data, grant
+   :author: Manuel
+   :location: BCN
+
+.. meta::
+   :description lang=en:
+      Collect and analyze project and build data to make better decision about our product
+      with a strong focus on the user experience.
```
agjohnson, line 9:
```suggestion
      Collect and analyze project and build data to make better decisions about our product
      with a strong focus on the user experience.
```

knowing-more-about-ourselves.rst
```diff
@@ -0,0 +1,206 @@
...
+or is generated by our own service.
+
+The data generated by Read the Docs which we already had access to was mostly high-level about our database objects.
+A few examples would be:
+
+* project configuration (e.g. PR build enabled, default branch)
+* build metadata (e.g. number of builds, length of builds)
+* platform information (redirects, webhooks, search queries, pageviews)
+
+With our new data collections project,
```
agjohnson, line 43:
```suggestion
With our new data collection project,
```

knowing-more-about-ourselves.rst
```diff
@@ -0,0 +1,206 @@
...
+   :align: center
+   :width: 100%
+
+
+This is a similar plot than the previous one but for `MyST parser <https://myst-parser.readthedocs.io/>`_.
+We are interested in knowing how many projects are writing Markdown using Sphinx.
+Many people don't know it's possible to write Markdown while keep using Sphinx and all the amazing features it has.
+Interestingly, the number of projects using MyST parser, currently, is higher than the number of projects using MkDocs.
+If you are reading this and you didn't know about MyST parser,
+we strongly recommend you to give it a try!
```
agjohnson, line 121: This is a good place for a link to a guide on using MyST

agjohnson (approved):
The narrative around each graph is a great addition. I think the readers will appreciate not just the graph, but the background on why each graph exists and the conclusions we drew from it :100:

knowing-more-about-ourselves.rst
humitos: I put the link at the beginning of this paragraph already. Should I link this last sentence to the "Get started" guide? https://myst-parser.readthedocs.io/en/latest/intro.html

knowing-more-about-ourselves.rst
humitos: This is done. I updated the plot with more accurate data.

knowing-more-about-ourselves.rst
```diff
@@ -117,6 +159,18 @@ Build time per project
...
+Thanks to this plot we were able to find some anomalies in the last weeks.
+We noticed that there were a few projects consuming 3x build time that their successor.
+This was due to the development workflow they follow (multiple pushes with small commits),
+making our platform to trigger multiple builds where each of them took more than 15 minutes to complete.
+This generated a bad UX, since they had to wait for *all the old builds to finish*
+before being able to see the results from the latest builds --  which was the only valid one.
+
+After noticing this happening pretty frequently to these projects,
+we prioritized the work on `Cancel old builds <https://github.com/readthedocs/readthedocs.org/issues/8961>`_,
+reducing the computing time for our servers and the waiting time for our users: Win-Win!
```
ericholscher, line 171:
```suggestion
Thanks to this plot we were able to find some anomalies in our builds.
We noticed that there were a few projects consuming 3x build time between different builds.
This was due to the development workflow they follow (multiple pushes with small commits),
making our platform to trigger multiple builds where each of them took more than 15 minutes to complete.
This caused a bad UX, since they had to wait for *all the old builds to finish*
before being able to see the results from the latest builds --  which was the only valid one.

After noticing this happening pretty frequently to these projects,
we prioritized the work on `Cancel old builds <https://github.com/readthedocs/readthedocs.org/issues/8961>`_,
reducing the computing time for our servers and the waiting time for our users: Win-Win!
```

ericholscher (approved):
Looks good, with 1 grammar update where it wasn't clear.

PR #184, opened 2022-10-14
sphinx-rtd-theme 1.1.0 release announcement

benjaoming:
Do not merge until the package is released :)

image


theme-release-110.rst
```diff
@@ -0,0 +1,77 @@
...
+    :tags: theme, release
+    :author: Benjamin
+    :location: Malmö, Sweden
+
+.. meta::
+    :description lang=en:
+        Information on sphinx-rtd-theme version 1.1.0
+
+
+Theme release 1.1.0
```
ericholscher, line 11:
```suggestion
Announcing sphinx_rtd_theme 1.1.0
```

This is really not obvious without context, we should be explicit both for SEO and for users.

theme-release-110.rst
```diff
@@ -0,0 +1,77 @@
...
+
+We are happy to announce the release of new version of our theme, `sphinx-rtd-theme`_. In this release, we have focused on bug fixes, backwards compatibility, and making the way for future releases.
+
+.. _sphinx-rtd-theme: https://sphinx-rtd-theme.readthedocs.io/en/stable/
+
+
+Changes and new features 💄
+---------------------------
+
+Visually, we have a couple of small tweaks that most people won't notice unless we mention them here 😇 The general strategy of the theme is to avoid breaking stuff, and that also goes for the visual parts.
```
ericholscher, line 22: The `breaking stuff` comment is a bit weird without context. I think we can frame it as maintaining backwards compatibility, in a more positive way.  

@benjaoming also 5th reminder to use semantic linebreaks I can comment on lines individually :)

theme-release-110.rst
```diff
@@ -0,0 +1,77 @@
...
+
+* The ``<kbd>`` tag now has a nicer styling.
+* Breadcrumb styling has been updated
+* Table cells with multiple paragraphs now have a correct formatting.
```
ericholscher, line 26: We should either end with a period, or not consistently.

theme-release-110.rst
```diff
@@ -0,0 +1,77 @@
...
+   :target: _images/sphinx-rtd-theme-110-screenshot3.png
+
+In the engine-room, we have ensured the long-term stability for users of the theme by putting upper bounds on ``Sphinx<6`` and ``docutils<0.18``. In the case of Sphinx, we know that the upcoming 6.x series (probably due later this year) will be incompatible with sphinx-rtd-theme 1.1 (we are working on fixes) and as for docutils 0.18, we already have the fixes lined up for version 1.2, but we need to drop compatibility with older Sphinx releases to include these.
+
+We also fixed an issue that caused the theme to fail when Sphinx ``5.2.0.post0`` was released and will ensure that this doesn't happen again.
+
+`Read the full changelog <changelog>`_
+
+
+.. _changelog: https://github.com/readthedocs/sphinx_rtd_theme/blob/master/docs/changelog.rst
```
ericholscher, line 50: Do we not render this changelog somewhere?

theme-release-110.rst
```diff
@@ -0,0 +1,77 @@
...
+
+
+How to upgrade
+--------------
+
+If you are using the theme for the first time, please refer the general `installation instructions <install>`_.
+
+For projects hosted on Read the Docs, the general update instruction is to modify your project's ``requirements.txt`` where you should add ``sphinx-rtd-theme==1.1.0`` (or replace any existing entries).
```
ericholscher, line 58: Is this only for RTD projects? Or all projects?

theme-release-110.rst
```diff
@@ -0,0 +1,77 @@
...
+
+If your project doesn't have a Python requirements file ``requirements.txt``, you need to add one in order to use newer versions of sphinx-rtd-theme. Read the Docs doesn't automatically update Python packages such as ``Sphinx`` or ``sphinx-rtd-theme`` because this risks breaking existing projects. You can read more about adding a ``requirements.txt`` in our :doc:`Documentation about Reproducible Builds <readthedocs:guides/reproducible-builds>`.
```
ericholscher, line 60: This is also not specifically true, and only for very old projects. I don't think we should include this in the post, as it's super specific, and we should just promote requirements.txt usage instead of documenting edge cases.

theme-release-110.rst
```diff
@@ -0,0 +1,77 @@
...
+-----------------
+
+Each little change comes with an overhead of testing, perfection and a long list of legacy support. We are addressing all that in upcoming releases, so it will become less cumbersome to add new features. The building and testing processes are refined and future releases will drop some of the legacy.
+
+Here are the highlights from our roadmap:
+
+* sphinx-rtd-theme 1.2: Adds docutils 0.18 support, possibly also docutils 0.19
+* sphinx-rtd-theme 2.0: Adds Sphinx 6.x support, dropping legacy support for several Sphinx releases and old browsers.
+
+If you wish to see more details, `view the full roadmap <roadmap>`_.
```
ericholscher, line 75: This post should have an ending. Asking the user to try out the theme, or give us feedback.

ericholscher (commented):
This looks good for the announcement. I have a few nitpicks to try and improve it.

theme-release-110.rst
```diff
@@ -0,0 +1,77 @@
...
+* The ``<kbd>`` tag now has a nicer styling.
+* Breadcrumb styling has been updated
+* Table cells with multiple paragraphs now have a correct formatting.
+* Definition lists rendered correctly in API docs
+* Issues with citation rendering and styling fixed
+* Long URLs break into several lines instead of overflowing
+
+.. image:: img/sphinx-rtd-theme-110-screenshot1.png
+   :width: 30%
+   :target: _images/sphinx-rtd-theme-110-screenshot1.png
```
ericholscher, line 33: This breaks in my testing -- is there a reason we have the `target` along with the path?

theme-release-110.rst
```diff
@@ -0,0 +1,77 @@
...
+Visually, we have a couple of small tweaks that most people won't notice unless we mention them here 😇 The general strategy of the theme is to avoid breaking stuff, and that also goes for the visual parts.
+
+* The ``<kbd>`` tag now has a nicer styling.
+* Breadcrumb styling has been updated
+* Table cells with multiple paragraphs now have a correct formatting.
+* Definition lists rendered correctly in API docs
+* Issues with citation rendering and styling fixed
+* Long URLs break into several lines instead of overflowing
+
+.. image:: img/sphinx-rtd-theme-110-screenshot1.png
```
humitos, line 31: I'd like to remove the browser frame from these images. I think it just add noise to the screenshot. Without it, it looks cleaner.

theme-release-110.rst
benjaoming: Agreed, I just copied from the former release announcement for consistency.

theme-release-110.rst
benjaoming: The URL is wrong

theme-release-110.rst
benjaoming: I wasn't sure what to do here, it works in the local build, whereas using `img/...` doesn't.

theme-release-110.rst
benjaoming: Agreed! should also redo the screenshots with neater annotations.

theme-release-110.rst
benjaoming: Thank you, you are absolutely right, and it was easy to update with a positive phrasing of the same thing.

As for semver.. yes, I try to upgrade my brain. I thought about writing a linter, but it's a bit hard to think up a _good_ syntax for something that's _semantic_ in its nature :)

theme-release-110.rst
benjaoming: Most definitely, fixed that :+1:

theme-release-110.rst
benjaoming: Updated to the stable URL... note that the current changelog is in the `latest` slug.

theme-release-110.rst
benjaoming: Good call, it's relevant for all projects. And I think you are also saying something important in that a project with a `requirements.txt` file might not "identify" as such and that will make the instructions look cluttered.

I'll try to think about something else. It should look simple.

benjaoming: Btw, the current text is updated... what I mean is that I would still look for simplifications.

theme-release-110.rst
benjaoming: Great idea! This is updated so it's simpler and just promotes using `requirements.txt` :+1:

theme-release-110.rst
benjaoming: Added a "Contributing" section, but it would be nice if you could look over it and add anything you'd like to see specifically for the theme... but maybe also if the blog post itself would have comments?

ericholscher (approved):
This looks better 👍

theme-release-110.rst
benjaoming: This was done by the way. I'd like to find a better recipe in the future.

theme-release-110.rst
benjaoming: Fixed with a slightly hack-ish replacement of `width` with `scale`

theme-release-110.rst
```diff
@@ -0,0 +1,91 @@
...
+   :align: right
+
+In the engine room, we have ensured the long-term stability for users of the theme by putting upper bounds on ``Sphinx<6`` and ``docutils<0.18``.
+
+We also fixed an issue that caused the theme to fail when Sphinx ``5.2.0.post0`` was released and will ensure that this doesn't happen again.
+
+`Read the full changelog <changelog>`_
+
+
+.. _changelog: https://sphinx-rtd-theme.readthedocs.io/en/stable/changelog.html
```
humitos, line 52: I don't think this is valid rST syntax. These links don't work.

theme-release-110.rst
```diff
@@ -0,0 +1,91 @@
...
+
+
+How to upgrade
+--------------
+
+If you are using the theme for the first time, please refer the general `installation instructions <install>`_.
```
humitos, line 58: Same here.

humitos: and there are others as well.

humitos (commented):
It seems there are some invalid links due a syntax error.

theme-release-110.rst
benjaoming: Thanks! This came into my muscle memory a long time ago, and I'm still try to get rid of it. I wish it worked :upside_down_face:

theme-release-110.rst
```diff
@@ -0,0 +1,81 @@
+.. post:: October 18, 2022
```
ericholscher, line 1:
```suggestion
.. post:: Nov 1, 2022
```

We've just announced the 1.1.0b3 on twitter. Let's wait a week and then plan to ship it on Nov 1?

PR #188, opened 2022-10-27
Build updates!

benjaoming:
This update also contains #169

* Update to latest Sphinx w/ docutils 0.17.1
* Update to latest ablog, lxml
* Remove transient dependencies from requirements.txt
* fail on warnings.
* Fix some broken intersphinx refs.
* Apply styles for several elements in docutils 0.17
* Shuffle around some margins, TBC in #191 

Before:

image

After:

image


humitos (approved):
LGTM! Not sure about the CSS changes, since it's not my topic, but they look 👍🏼 

cc'ing @agjohnson just in case he wants to review it.

benjaoming:
What this PR mainly changes is margins, but it also re-aligns "Recent posts" with contents below it, which I think is nicer for the eye.

I've opened https://github.com/readthedocs/blog/issues/191 to re-address margins.

benjaoming:
I've been clicking around on a lot of blog posts now and stopped noticing things that look wrong.

So I think we're good to ship this.

It's supporting docutils 0.17 btw.

_themes/rtd-blog/static/local.css
```diff
@@ -86,25 +95,45 @@ div.admonition {
...
+
+dl.footnote dt {
+  float: left;
+  margin-left: 1rem;
+  margin-right: 1rem;
+  margin-top: 0;  
+}
+
+dl.footnote dd > p:first-child {
+  margin: 0px;
```
benjaoming, line 113: On second thought, using floats isn't nice. Should probably be `display: table` or flexbox

agjohnson:
Just noting that we don't plan to continue using our blog here. Our blog will be contained in our website repository, so spending much time here fixing things up is not a priority.

We don't have a timeframe, and need to consider if/how we port content:

https://github.com/readthedocs/site-community/issues/33

benjaoming:
Ah yes, that's much better than the current blog. I wasn't aware. I'll wrap up the footnotes issue :+1: 

Love the issue description over in https://github.com/readthedocs/site-community/issues/33 :heart:

_themes/rtd-blog/static/local.css
ericholscher: I think in general we should avoid doing any CSS changes here. It's not work we're going to get value from in the long term.

agjohnson (approved):
LGTM

benjaoming:
Replaced the floating CSS in footnotes with `display: table`, looks fine in Chrome and Firefox.

image

_themes/rtd-blog/static/local.css
benjaoming: Used `display: table` in order to not wrestle with flexbox :)

PR #193, opened 2022-11-08
Enable fail on warning to avoid shipping broken links

ericholscher:


agjohnson (approved):
Change is :+1: but checks are grumpy now of course

ericholscher:
Replaced by #188

PR #196, opened 2022-11-10
Revert #188

benjaoming:
This reverts changes in #188 because of reversed atom feed order.

Note: We should probably still keep an eye on the newsletter service since reversing the RSS might make it send out the same newsletter once again :smiling_face_with_tear:


benjaoming:
We opted for the workaround in #197 

There is an upstream fix coming.

PR #197, opened 2022-11-11
Limit RSS feed to 1 post to disable sending old posts

ericholscher:


ericholscher:
👍 https://readthedocs-blog--197.org.readthedocs.build/archive/atom.xml

benjaoming (approved):
This is good! Would still keep an extra eye on what newsletter is about to be sent out.

It's also nice because we can look forwards and not do the revert of #188 while aBlog might get a fix or option to reverse the RSS order.

PR #209, opened 2023-02-16
Read the Docs ❤️ Django

benjaoming:
First draft

### DIFF
--- a/we-love-django.rst
+++ b/we-love-django.rst
@@ -1,0 +1,58 @@
+.. post:: Jan 26, 2023
+   :tags: newsletter, python
+   :author: Ben
+   :location: MLM
+
+.. meta::
+   :description lang=en:
+      Read the Docs has a special relationship with Django.
+
+Read the Docs ❤️s Django
+========================
+
+tldr;
+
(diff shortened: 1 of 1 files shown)


ericholscher:
This seems like a good start, but it feels very fluffy, and buries the lede. I think we want to highlight the sponsorship at the top of the post? 

I also feel like the first section doesn't really say anything? I can't really summarize it other than "We think people choose Django because it has good documentation, but we have no proof other than our core team uses it sometimes"?  The stuff about the core team doing OSS contribution also feels off topic here, and doesn't really drive to a specific point that seems relevant to this post from what I can tell?

I think overall we should cut down the length, and focus on a few key points:

* We are sponsoring Django
* Django integrates docs into the product process, and that's an inspiration
* RTD was inspired by Django, and we want to contribute back. We also contribute back in many other ways (Community ads, hosting their PDF docs and a mirror, etc.)

benjaoming:
I think it's best if you implement the feedback here @ericholscher. I don't think the blog post is far from ready, but I can't quite guess how to implement the feedback. Please feel free to sign off and wrap up this blog post :+1:

humitos:
Should we publish or close this?

PR #217, opened 2023-05-12
Shorten some "wall of text" issues in newsletter

benjaoming:
I saw a preview of the newsletter and didn't like the text amount.

I trimmed it a bit and de-prioritized some text from the first item.


benjaoming:
Going to merge this so it's active when the newsletter is sent out.

PR #221, opened 2023-05-31
Add explicit date

ericholscher:


migrate-configuration-v2.rst
```diff
@@ -15,7 +15,7 @@ for all projects in order to build documentation successfully.
 We will stop creating a default configuration file *behind the scenes*, making assumptions for your projects,
 and automatically fixing problems with project configurations.
 
-We plan to deploy these changes and **start failing builds without a configuration file in September 2023**.
+We plan to deploy these changes and **start failing builds without a configuration file on September 20, 2023 at 12:01am UTC**.
```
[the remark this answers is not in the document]
benjaoming: A quick note: We can use "brownout" internally, but it's not gonna make much sense in a public announcement :)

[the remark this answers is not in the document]
agjohnson: Brownout is a term fairly commonly used here, but probably worth using some supporting language here as well. Here is a GitHub announcement for comparison:
https://github.blog/changelog/2021-08-10-brownout-notice-api-authentication-via-query-parameters-for-48-hours/

Also noting, this was the third brownout and it lasted 48h. I wonder if instead of 2h blocks we instead just do something like a 12h period (July 24) and a 24h period (Aug 7).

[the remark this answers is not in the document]
benjaoming: I see! allow me to revise that: We cannot use brownout without accompanying it with a good description of what it means :)

benjaoming (approved):
In order to have the date visible quickly, I'm fine about merging this and adding other dates in a second iteration.

But if someone adds the brownout dates (+ nice with a definition of what we mean by that) in this PR, that's great, too :+1:

migrate-configuration-v2.rst
```diff
@@ -7,26 +7,36 @@
...
 which is the OpenSSL version installed by the build image used when the project doesn't specify a configuration file.
 
 The main reason behind this decision is that it's hard to maintain a set of defaults that are useful
-for all of our users and the different tools and libraries they use to build documentation.
-Without a configuration file, we must guess at the configuration the project needs to build and this is also difficult to do reliably.
+for all of our users and the different tools they use to build documentation.
+Without a configuration file, we must guess the configuration the project needs to build, which is difficult to do reliably.
 Even if the defaults we chose worked well for a period of time,
-we found them impossible to change over time.
+we found it impossible to change over time.
```
agjohnson, line 26: "It" is in wrong agreement here ("defaults")

```suggestion
we found them impossible to change over time.
```

agjohnson: Or

```suggestion
we found it impossible to change them over time.
```

migrate-configuration-v2.rst
```diff
@@ -7,26 +7,36 @@
...
 Doing so would break projects that were relying on our defaults instead of specifying a configuration file
 and pinning their dependencies (e.g. OS, Python version, dependencies, etc).
 
+Deprecation timeline
+--------------------
+
+We understand this change will effect many of our users,
```
agjohnson, line 33:
```suggestion
We understand this change will affect many of our users,
```

migrate-configuration-v2.rst
```diff
@@ -7,26 +7,36 @@
...
+so we have a timeline to communicate this deprecation to our users effectively.
+
+* **June 2023**: Email project owners with unsupported configurations once a week, along with an onsite notification to alert them of the deprecation.
+* **Monday, July 24, 2023**: Do the first brownout (temporarily enforce this deprecation) for 2 hours in 3 timezones, 9:00-11:00, in 3 timezones: UTC-8, UTC, and UTC+8.  
+* **Monday, August 7, 2023**: Do a second brownout (temporarily enforce this deprecation) for 2 hours in 3 timezones, 9:00-11:00, in 3 timezones: UTC-8, UTC, and UTC+8.
```
agjohnson, line 38: Brought up in comments here, but maybe lets make this even easier on us and replicate what GitHub did: 12h window first date, 24h window the second date. I'm guessing it might be hard to have folks around for all of these 2h windows, thinking about this a bit more. We'll definitely catch more projects this way too.

migrate-configuration-v2.rst
```diff
@@ -7,26 +7,36 @@
 Migrate your project to ``.readthedocs.yaml`` configuration file v2
 ===================================================================
 
-We are announcing the **deprecation of building without a configuration file** (``.readtehdocs.yaml``)
-together with the deprecation of configuration file v1 as well.
+We are announcing a new **requirement for all builds to use our configuration file version 2: ``.readthedocs.yaml``**.
```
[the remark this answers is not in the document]
benjaoming: Yes, the filename is in the headline right above, so that's fine :+1:

migrate-configuration-v2.rst
agjohnson: If we start a 12h window Mon Sept 4 @ 0:00 PT and go till 12:00 PT, this is what the overlap looks like:

image

Asia users will get errors starting at the end of their day and the errors will resolve overnight, so not *great*, but we have more users in EU/US anyways.

ericholscher: Sounds good to me 👍

migrate-configuration-v2.rst
ericholscher: I always mess this one up 🙃

agjohnson:
Note, before merging this, urllib3 might not be planning to add a deprecation exception like we saw in urllib3 2.0. This would allow us to relax the timeframe a little bit, and maybe aim for end of Sept instead. Still waiting on confirmation here, in #222

agjohnson:
Confirmed that urllib3 *will not* add a deprecation exception like urllib3 2.0 did. Because of this we will not have to worry about builds failing due to outdated libssl come Sept 11.

So, we can now set a date later in Sept (or even call it the first week of Oct) to give a bit more room for this migration timeframe.

We also don't need to be strict about the deprecation date now either. If projects aren't migrating over, we had talked about allowing some projects to continue building after the cutoff (big or long standing projects having trouble with the migration).

ericholscher:
@agjohnson I guess I don't understand the reasoning for changing the dates we agreed to on the call? Is there a specific worry you're trying to address?

agjohnson:
Just gaining 3 weeks on our timeframe. We described the timeframe as fairly tight. We could rather quickly decide on Monday Sept 25 and keep everything else the same.

ericholscher:
I'm fine with that date, and gives us room for a 3rd brownout if we need it.

agjohnson:
That's a great idea. We could add another 24h or even 48h brownout on Sept 4 in place of the cutover.

Looks like everyone is planning on being around Mon Sept 25 too, just in case. We will have been through this 3 times at that point though, so I doubt there will be many fires then.

migrate-configuration-v2.rst
```diff
@@ -7,26 +7,36 @@
...
-Read the Docs *will soon start requiring* a ``.readthedocs.yaml`` configuration file
+Read the Docs *will start requiring* a ``.readthedocs.yaml`` configuration file
 for all projects in order to build documentation successfully.
-We will stop creating a default configuration file *behind the scenes*, making assumptions for your projects,
-and automatically fixing problems with project configurations.
+We will stop supporting builds without explicit configuration,
+because this creates implicit dependencies that users aren't aware of.
 
-We plan to deploy these changes and **start failing builds without a configuration file in September 2023**.
+**We plan to start failing builds not using configuration file version 2 on September 4, 2023**.
```
agjohnson, line 18:
```suggestion
**We plan to start failing builds not using configuration file version 2 on September 25, 2023**.
```

migrate-configuration-v2.rst
```diff
@@ -7,26 +7,36 @@
...
+Deprecation timeline
+--------------------
+
+We understand this change will affect many of our users,
+so we have a timeline to communicate this deprecation to our users effectively.
+
+* **June 2023**: Email project owners with unsupported configurations once a week, along with an onsite notification to alert them of the deprecation.
+* **Monday, July 24, 2023**: Do the first brownout (temporarily enforce this deprecation) for 12 hours: 00:01 PST to 11:59 PST (noon)
+* **Monday, August 7, 2023**: Do a second brownout (temporarily enforce this deprecation) for 24 hours: 00:01 PST to 23:59 PST (midnight)
+* **Monday, September 4, 2023**: Fully remove support for building documentation without configuration file v2.
```
agjohnson, line 39: Adding a third brownout, evenly space brownouts at 3 weeks apart (Aug 7 brownout is Aug 14 brownout). Team is available on all these dates, based on our shared calendar.

```suggestion
* **Monday, July 24, 2023**: Do the first brownout (temporarily enforce this deprecation) for 12 hours: 00:01 PST to 11:59 PST (noon)
* **Monday, August 14, 2023**: Do a second brownout (temporarily enforce this deprecation) for 24 hours: 00:01 PST to 23:59 PST (midnight)
* **Monday, September 4, 2023**: Do a third and final brownout (temporarily enforce this deprecation) for 48 hours: 00:01 PST to 23:59 PST (midnight)
* **Monday, September 25, 2023**: Fully remove support for building documentation without configuration file v2.
```

agjohnson (approved):
Two changes:

- Bump brownout Aug 7 -> Aug 14 to evenly space
- Sep 4 brownout for 48h added

If 24h seems more right, lets go with that though.

humitos (approved):
This looks good to me with Anthony's suggestions 👍

PR #226, opened 2023-07-06
Post: small post about "latest core requirements"

humitos:
Announce the changes made at
https://github.com/readthedocs/readthedocs.org/pull/10508/

Note that we are not exposing this feature flag to our users. We don't want them to contact us to enable this flag. If they need to install the latest versions they can just specify them in their `requirements.txt` file.

### DIFF
--- a/python-core-requirements-changed.rst
+++ b/python-core-requirements-changed.rst
@@ -1,0 +1,25 @@
+.. post:: July 7, 2023
+   :tags: builders
+   :author: Manuel
+   :location: BCN
+   :category: Changelog
+
+Python "core requirements" installed by default have changed
+============================================================
+
+Starting on **August 7, 2023** all new projects imported on Read the Docs
+will install only ``sphinx``, ``mkdocs`` and ``readthedocs-sphinx-ext`` as "core requirements".
+Besides, the new behavior will install the latest version of these requirements.
+
+Note that previously Read the Docs was installing also
+``jinja``, ``sphinx-rtd-theme``, ``pillow``, ``mock``, ``alabaster``, ``commonmark`` and ``recommonmark``
+specifying particular versions depending on different factors that were confusing for users and hard to debug.
+
+We are moving away from executing commands and installing dependencies on behalf of the users
+and trying to make the build experience on Read the Docs the same as building locally,
+without surprises.
+
+Projects imported previous to August 7, 2023 will keep the actual behavior and won't be affected by this change.
+If you have a project you want to install the latest version of these requirements,
+you can use a ``requirements.txt`` file to specify their versions.
+Read more about :doc:`how to use a requirements.txt file in our documentation <readthedocs:config-file/v2>`.
--- a/python-core-requirements-changed.rst
+++ b/python-core-requirements-changed.rst
@@ -4,12 +4,12 @@
    :location: BCN
    :category: Changelog
 
-Python "core requirements" installed by default have changed
-============================================================
+Python "core requirements" for new projects will install latest version
+=======================================================================
 
 Starting on **August 7, 2023** all new projects imported on Read the Docs
 will install only ``sphinx``, ``mkdocs`` and ``readthedocs-sphinx-ext`` as "core requirements".
-Besides, the new behavior will install the latest version of these requirements.
+The default behavior will be to install the latest version of these requirements.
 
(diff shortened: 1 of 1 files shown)


python-core-requirements-changed.rst
```diff
@@ -0,0 +1,25 @@
...
+specifying particular versions depending on different factors that were confusing for users and hard to debug.
+
+We are moving away from executing commands and installing dependencies on behalf of the users
+and trying to make the build experience on Read the Docs the same as building locally,
+without surprises.
+
+Projects imported previous to August 7, 2023 will keep the actual behavior and won't be affected by this change.
+If you have a project you want to install the latest version of these requirements,
+you can use a ``requirements.txt`` file to specify their versions.
+Read more about :doc:`how to use a requirements.txt file in our documentation <readthedocs:config-file/v2>`.
```
ericholscher, line 25:
```suggestion
.. post:: July 7, 2023
   :tags: builders
   :author: Manuel
   :location: BCN
   :category: Changelog

Python "core requirements" for new projects will install latest version
=======================================================================

Starting on **August 7, 2023** all new projects imported on Read the Docs
will install only ``sphinx``, ``mkdocs`` and ``readthedocs-sphinx-ext`` as "core requirements".
The default behavior will be to install the latest version of these requirements.

Note that previously Read the Docs was installing also
``jinja``, ``sphinx-rtd-theme``, ``pillow``, ``mock``, ``alabaster``, ``commonmark`` and ``recommonmark``
specifying particular versions depending on different factors that were confusing for users and hard to debug.

We are moving away from executing commands and installing dependencies on behalf of the users
and trying to make the build experience on Read the Docs the same as building locally,
without surprises.

Projects imported before August 7, 2023 will keep the existing behavior and won't be affected by this change.
If you have a project you want to install the latest version of these requirements,
you can use a ``requirements.txt`` file to specify their versions, or no version to automatically install the latest version.
Read more about :doc:`how to use a requirements.txt file in our documentation <readthedocs:config-file/v2>`.
```

ericholscher (approved):

python-core-requirements-changed.rst
humitos: > If you have a project you want to install the latest version of these requirements,
you can use a ``requirements.txt`` file to specify their versions, or no version to automatically install the latest version.

I thought about this as well, but this is not true:

```
$ pip install "sphinx<2.2"
...
$ echo "sphinx" > requirements.txt
$ pip install -r requirements.txt
```

Nothing was upgraded at this point. However, projects with `PIP_ALWAYS_UPGRADE` will add `--upgrade` to that command:

```
$ pip install --upgrade -r requirements.txt
```

and the latest Sphinx will be installed.

humitos: This code lives at https://github.com/readthedocs/readthedocs.org/blob/78751790cdecd22c635611b51b35779022a0ccd4/readthedocs/doc_builder/python_environments.py#L296-L297

We only have 5 projects on .org with this feature flag enabled (https://admin.ops.readthedocs.org:10000/admin/projects/feature/5/change/). Does it make sense to apply it for all the projects? Man, we have a lot of feature flags that make our workflow too complex 😞 

I think it makes sense, because it's aligned with what the user wants, right? If they specify `sphinx` without versions it's because they want the latest one.

humitos: On .com we don't even have that feature flag created, so, 0 projects are using it.

humitos: Pip documentation about `--upgrade`: https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-U

ericholscher: It sounds like users can set the ENV var to have pip always upgrade? If that works, I'm fine with that until users can call pip with whatever flags they want in `build.jobs.install`.

humitos: That's a pretty good idea actually. I wasn't aware this can be done with environment variables. I think it would be really good to document/suggest this for projects with this requirement 👍

python-core-requirements-changed.rst
```diff
@@ -0,0 +1,25 @@
+.. post:: July 7, 2023
```
humitos, line 1:
```suggestion
.. post:: July 10, 2023
```

PR #231, opened 2023-07-19
Copy atom_absolute extension from WTD and rewrite multiple categories

ericholscher:


exts/atom_absolute.py
```diff
@@ -7,49 +7,56 @@
...
+    rewritten_feed_path = os.path.join(
+        app.outdir, 'blog/archive/category/{tag}/atom_absolute.xml'.format(tag=tag)
+    )
 
-        # Rewrite the namespace map to not be Null
-        namespace = list(root.nsmap.values())[0]
-        nsmap = {'atom': namespace}
+    if not os.path.isfile(feed_path):
+        logger.error('Atom feed does not exist at: {}'.format(feed_path))
+        return
```
humitos, line 40: Shouldn't we raise an exception here so the build fails and we notice it?

exts/atom_absolute.py
```diff
@@ -7,49 +7,56 @@
...
+        html_content = html.fromstring(content.text)
+        # Remove janky system messages
+        for to_remove in html_content.find_class('system-message'):
+            to_remove.drop_tree()
+        # fix links
+        html_content.make_links_absolute(feed_url)
+        content.text = html.tostring(html_content)
 
-def setup(app):
-    app.connect('build-finished', rewrite_atom_feed)
```
humitos, line 49: Don't we need to keep this as an extension?

humitos (commented):
I'm not sure to understand how this works. Is the idea behind this is to remove `newsletter` and `changelog` entries from the XML file?

ericholscher:
This was harder to get working than I hoped, so gonna hold off for now.

PR #249, opened 2023-11-16
Post: announcing versioning schemes

stsewd:
ref:

- https://github.com/readthedocs/readthedocs.org/pull/10845
- https://github.com/readthedocs/readthedocs.org/pull/10766


introducing-versioning-schemes.rst
```diff
@@ -0,0 +1,51 @@
+.. post:: November 28, 2023
+   :tags: docs, versions
+   :category: Changelog
+   :author: Santos
+   :location: CUE
+
+Introducing versioning schemes
+==============================
```
ericholscher, line 8:
```suggestion
Introducing versioning schemes
==============================
```
```suggestion
Introducing support for version-only projects
=============================================
```

introducing-versioning-schemes.rst
```diff
@@ -0,0 +1,51 @@
...
+  Resulting in URLs like:
+
+  - ``/index.html``
+  - ``/api/v3.html``
+
+While not all projects have or require translations, many benefit from having multiple versions.
+This meant that projects without translations but with need for multiple versions,
+had to opt into supporting translations in order to have multiple versions.
+
+We're excited to announce that we now support projects with multiple versions without translations.
```
ericholscher, line 33:
```suggestion
**We're excited to announce that we now support projects with multiple versions without translations.**
```

ericholscher (approved):

introducing-versioning-schemes.rst
stsewd: what about "Introducing support for projects with multiple versions without translations"? It's a long title, not sure which one is better p:

ericholscher: I think we can keep it as suggested -- it's the thing that's changing, so seems good to highlight. 👍

PR #251, opened 2023-12-07
Post: announcing the new improvements to redirects

stsewd:
Ref https://github.com/readthedocs/readthedocs.org/pull/10881

Preview at https://readthedocs-blog--251.org.readthedocs.build/new-improvements-to-redirects/

### DIFF
--- a/new-improvements-to-redirects.rst
+++ b/new-improvements-to-redirects.rst
@@ -1,4 +1,4 @@
-.. post:: January ?, 2024
+.. post:: January 2, 2024
    :tags: redirects, docs
    :category: Changelog
    :author: Santos
--- a/new-improvements-to-redirects.rst
+++ b/new-improvements-to-redirects.rst
@@ -1,0 +1,75 @@
+.. post:: January ?, 2024
+   :tags: redirects, docs
+   :category: Changelog
+   :author: Santos
+   :location: CUE
+
+New improvements to redirects
+=============================
+
+As your documentation evolves, content is moved, and pages are renamed,
+leading to broken links for your users.
+Redirects allow you to point users to the new location of a page.
+
+We are excited to announce significant improvements to our redirects feature to make them more flexible and powerful.
+
+Key improvements
+----------------
+
+HTTP status code
+  You can now choose the `HTTP status code`_ for your redirects,
+  selecting between 301 (permanent redirect) and 302 (temporary redirect).
+  Previously, all redirects had the status code 302.
+
+.. _HTTP status code: https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections
+
+Explicit ordering
+  Redirects now have an explicit order,
+  allowing you to control the order in which they are processed.
+  Previously, redirects were ordered based on their last modification date.
+
+Disable redirects without deletion
+   You can now disable redirects without deleting them.
+   This is useful when debugging or when you want to temporarily disable a redirect.
+
+Trailing slash
+  Redirects now apply to URLs with and without a trailing slash.
+  Previously, you had to create two redirects to handle both cases.
+
+Suffix wildcards
+  While we already supported suffix wildcard redirects, their functionality was limited.
+  Now wildcard redirects use a more familiar syntax, ``*`` instead of ``$rest``,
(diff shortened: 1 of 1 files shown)


new-improvements-to-redirects.rst
```diff
@@ -0,0 +1,75 @@
...
+  Use ``*`` instead.
+- When using a wildcard, the matched suffix is no longer added to the target URL automatically.
+  Use the ``:splat`` placeholder in the target URL to add the matched suffix.
+- The redirect type ``sphinx_html`` used in the API was renamed to ``clean_url_to_html``.
+- The redirect type ``sphinx_htmldir`` used in the API was renamed to ``html_to_clean_url``.
+- The prefix redirect type was removed.
+  Use an *Exact* redirect with a wildcard instead.
+
+If you had redirects using the old syntax, you don't need to do anything.
+We have automatically migrated them to use the new syntax.
```
humitos, line 75: Since everything is automated and the user doesn't need to do anything, I wouldn't call this section "Breaking changes" to avoid calling their attention and thinking their redirects are currently breaking.

new-improvements-to-redirects.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Trailing slash
+  Redirects now apply to URLs with and without a trailing slash.
+  Previously, you had to create two redirects to handle both cases.
+
+Suffix wildcards
+  While we already supported suffix wildcard redirects, their functionality was limited.
+  Now wildcard redirects use a more familiar syntax, ``*`` instead of ``$rest``,
+  and the ``:splat`` placeholder is available in the target URL.
+  Wildcards can be used with *Exact* and *Page* redirects.
```
ericholscher, line 43: This could use a link to the docs on this, I think.

new-improvements-to-redirects.rst
```diff
@@ -0,0 +1,75 @@
...
+
+.. _our documentation: https://docs.readthedocs.io/en/stable/user-defined-redirects.html#user-defined-redirects
+
+Breaking changes
+----------------
+
+In order to support these new improvements, we have made some breaking changes:
+
+- The old ``$rest`` syntax for wildcard redirects is no longer supported.
+  Use ``*`` instead.
```
ericholscher, line 66: Are we warning users in the form if they try to use `$rest`? Probably a good idea if we're not, since users won't all  read this post.

new-improvements-to-redirects.rst
ericholscher: I don't have a better word for it, so I think breaking changes are fine. It will break API integrations for example, so probably good to alert folks.

ericholscher (approved):
This is a great post 💯

new-improvements-to-redirects.rst
```diff
@@ -0,0 +1,75 @@
...
+  You can now choose the `HTTP status code`_ for your redirects,
+  selecting between 301 (permanent redirect) and 302 (temporary redirect).
+  Previously, all redirects had the status code 302.
+
+.. _HTTP status code: https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections
+
+Explicit ordering
+  Redirects now have an explicit order,
+  allowing you to control the order in which they are processed.
+  Previously, redirects were ordered based on their last modification date.
```
ericholscher, line 29: Did we ever make a change to Forced redirects with this work? I know we discussed something like:

* Always execute Exact redirects
* Only allow wildcard redirects on paid plans, since they are the more expensive ones to query

I think if possible getting rid of the Forced concept would be another really big win with this announcement, if we could do it.

ericholscher: Edit: I see in the PR now that it's an open question... will address it here :)

new-improvements-to-redirects.rst
ericholscher: Found this in the PR as well :D

new-improvements-to-redirects.rst
```diff
@@ -0,0 +1,75 @@
+.. post:: January 2, 2024
```
stsewd, line 1:
```suggestion
.. post:: January 3, 2024
```

